PR #7, opened 2017-04-24
Add bisection root-finder in arbitrary dimension

dpsanders:
This adds a bisection root finder that works with single or multiple variables.

Currently, this involves modifying the `Root` type.
For bisection in particular, this is, in some sense, not necessary, since it cannot prove uniqueness, so any roots are `:unknown`.

However, for multi-dimensional root finding in general, it is maybe necessary to do so.


dpsanders:
Moved from https://github.com/JuliaIntervals/ValidatedNumerics.jl/pull/262

dpsanders:
Any comments, @lbenet?

dpsanders:
If you don't have any problem, I'm planning to merge this @lbenet.

lbenet:
I will review it now; if I don't come back to you before lunch time, simple go ahead.... Sorry.

examples/complex_roots.jl
```diff
@@ -0,0 +1,50 @@
...
+"""
+function complex_bisection(f, X::IntervalBox)
+
+    """
+    Make a 2D real version of the complex function `f` suitable for `bisection`,
+    i.e. that accepts an `IntervalBox` and returns an `IntervalBox`
+    """
+    function g(X::IntervalBox)
+        x, y = X
+        z = x + im*y;
```
lbenet: Not really transcendental, but I read elsewhere that this is better written as `z = complex(x, y)` (avoids one multiplication and one addition).

lbenet: Actually, it maybe even better as `z = complex( X... )` directly.

dpsanders: Good point, thanks.

dpsanders:
I would like to add `complex_bisection` as part of the source and export it.
(It's currently just in `examples`.)

It's not the best name, but I can't think of a better one right now, and it's pretty nice that it's so easy to write.

lbenet:
> I would like to add complex_bisection as part of the source and export it.
(It's currently just in examples.)

I agree with you.

lbenet:
LGTM. I'll wait to travis before merging.

test/bisection.jl
```diff
@@ -0,0 +1,34 @@
+using ValidatedNumerics, ValidatedNumerics.RootFinding
```
dpsanders: Good catch, thanks.

PR #8, opened 2017-04-28
Multidimensional Newton operator

dpsanders:
Add a multi-dimensional Newton operator.


dpsanders:
Note that this requires changes to the `@intervalbox` macro.
That macro should probably be moved from IntervalArithmetic.jl to this package.

dpsanders:
If you have time to look at this one too, that would be great @lbenet. 

As far as I remember, this follows on from the bisection one and so that should be merged first.

lbenet:
I'll take a look on this too, but hold me on about 30 min. A quick remark: tests are not passing.

lbenet:
> That macro should probably be moved from `IntervalArithmetic.jl` to this package.

But then `IntervalArithmetic.jl` would be only of use for one-dimensional intervals. I do think that also the multidimensional definitions should be in `IntervalArithmetic.jl`, while applications should build upon that. Or am I missing something?

src/multidim/intervalbox.jl
```diff
@@ -0,0 +1,55 @@
+# This file is part of the ValidatedNumerics.jl package; MIT licensed
```
lbenet: I think this file should be in IntervalArithmetic.jl; see [this comment](https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/8#issuecomment-299942802)...

src/multidim/intervalbox_macro.jl
```diff
@@ -0,0 +1,56 @@
+using ValidatedNumerics
```
lbenet: I think this file should belong to `IntervalArithmetic.jl"; see above.

dpsanders:
I am not suggesting removing `IntervalBox` from `IntervalArithmetic.jl` -- I completely agree with you that it belongs there.

I am only talking about the `@intervalbox` macro that is wrapped around a function in order to automatically define the standard function that takes two variables (I think it is only for 2 variables currently?), as well as versions that act on `IntervalBox`es  and `Array`s.

Maybe this is, however, not the correct solution, and rather a new function should be defined like I did for `complex_bisection`.

src/newton.jl
```diff
@@ -18,14 +18,35 @@ function guarded_mid{T}(f, x::Interval{T})
...
+    J = ForwardDiff.jacobian(f, [X...])
+
+    # @show m
+    # @show J
+
+    return IntervalBox( (m - (J \ f(m))... ) )
+end
+
 
 doc"""If a root is known to be inside an interval,
```
lbenet: Shouldn't there be a `newtow_refine` method for `IntervalBox`s?

lbenet:
Regarding the necessity (or not) of `@intervalbox`, I'd like to point out that this already works!

```julia
julia> using ValidatedNumerics

julia> X = IntervalBox(1..1, 2..2)
[1, 1] × [2, 2]

julia> f(x, y) = (x + y, x - y)
f (generic function with 1 method)

julia> f(X...)
([3, 3],[-1, -1])

```

Maybe my concern with `@intervalbox` is that the name reminds me `@interval`, so I don't relate it with adpating a function, but with the creation of an `IntervalBox`.

dpsanders:
The difference is that `@intervalbox` defines a version of `f` such that `f(X)` takes an `IntervalBox` and **returns** an `IntervalBox`. I seem to have broken my VN installation though, so can't show it right now. (`@intervalbox` is not defined?)

dpsanders:
And to use `ForwardDiff` for automatic differentiation, as in this PR, we need a version that takes an `Array` as argument. It's all a bit of a mess.

lbenet:
> The difference is that `@intervalbox` defines a version of `f` such that `f(X)` takes an `IntervalBox` and returns an `IntervalBox`.  I seem to have broken my VN installation though, so can't show it right now. (`@intervalbox` is not defined?)

Sorry, I missed that. Still, I dislike the name though, because I think of a shortcut similar to `@interval`. In any case, it may be a good idea to have a macro of that kind, or as you proposed, perhaps a function.

In any case, the contests of `src/multidim/intervalbox.jl` include the definition of `IntervalBox`, as well as other general methods; that's what I was criticizing.

dpsanders:
Yes, the name is not great.

Ah yes, oops.

lbenet:
I think this could be merged, though it would be nice to add some tests.

dpsanders:
Closing in favour of #24.

PR #12, opened 2017-05-13
Remove tests for promotion to Dual

dpsanders:

PR #15, opened 2017-05-19
Add promotion_rule to handle promotion involving Dual

lbenet:
Related to https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/13

### DIFF
--- a/src/IntervalRootFinding.jl
+++ b/src/IntervalRootFinding.jl
@@ -17,10 +17,18 @@ export
     bisection,
     bisect
 
-import Base: ⊆, show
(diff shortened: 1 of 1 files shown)


lbenet:
This is an alternative to #12...

dpsanders:
Are these actually necessary, though? The package seems to work fine without any of these promotion rules.

lbenet:
I don't recall why we needed them at all. So I'll close this and we check #12.

dpsanders:
I think we needed them because of some ambiguities on Julia 0.4. But it seems they were only "potential" ambiguities that were never actually encountered!

PR #16, opened 2017-05-19
Add Intervalbox macro

dpsanders:
Moved from IntervalArithmetic.


dpsanders:
I don't think this approach is very fruitful. Closing.

PR #17, opened 2017-05-23
use version number instead of release in .travis.yml

tkelman:
release will change over time, but your REQUIRE file says this package supports julia 0.5 so it should continue to be tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ os:
   - osx
 
 julia:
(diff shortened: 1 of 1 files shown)


lbenet:
I think this can be merged right away. Another alternative is to test 0.6 and drop off 0.5. 

@dpsanders, what do you think?

dpsanders:
Done in #24.

PR #18, opened 2017-05-28
Generic branch-and-prune algorithm

dpsanders:


dpsanders:
This is an initial attempt to rewrite Krawczyk, Newton and bisection in terms of a single, generic branch-and-prune routine. 

So far, it works only for 1D functions, since it is not clear how to extract the dimension of the input IntervalBox and pass it through correctly. 

It also currently ignores the cleverness of the Newton method when the derivative over the interval contains 0, but that should be relatively easy to fix, e.g. by allowing the pruning functions to return a vector of intervals instead of a single interval.

cc @lbenet

dpsanders:
Note that currently it is restricted to searching for roots of the provided function.

dpsanders:
Closing in favour of #24 .

PR #24, opened 2017-06-30
Complete rewrite using generic branch and prune with contractors

dpsanders:


src/branch_and_prune.jl
```diff
@@ -0,0 +1,205 @@
...
+        NX =  refine(X -> C.op(C.f, C.fp, X), NX)
+        return :unique, NX
+    end
+
+
+    return :unknown, NX
+end
+
+
+roots{C<:Contractor}(f, X, contractor::Type{C}, tol::Float64=1e-3) = branch_and_prune(f, X, contractor, tol)
```
lbenet: Add docstrings for `roots`

lbenet:
I have done some work here for tests to pass. Essentially some functions are currently unexported, among other minor things. Shall I commit directly to this branch, or shall I do it in an alternative one?

dpsanders:
Just add it to this huge branch I think. Thanks!

lbenet:
Just pushed a PR which allows to pass tests locally. 

If you can take a look on the changes it would be worth doing it, because I think your plan was to deprecate/erase some of the previous functions (e.g., use `roots` instead of `newton` or `krawczyk`).

lbenet:
Done; tests are passing...

dpsanders:
Thanks for the tests! I have added a basic docstring for roots.

I think we should merge this and refine later!

dpsanders:
I converted some of examples.jl into new tests, which pass. 
Merging this for now!

PR #27, opened 2017-12-08
Add Interval as constant

lbenet:
Tests are failing after cloning because `Interval` is not being recognized. I simply solve this by adding it as a `const`

### DIFF
--- a/src/IntervalRootFinding.jl
+++ b/src/IntervalRootFinding.jl
@@ -21,6 +21,7 @@ import Base: ⊆, show
 
 const derivative = ForwardDiff.derivative
 const D = derivative
+const Interval = IntervalArithmetic.Interval
 
 # Root object:
 immutable Root{T<:Union{Interval,IntervalBox}}
--- a/test/bisect.jl
+++ b/test/bisect.jl
@@ -1,11 +1,12 @@
 using IntervalArithmetic, IntervalRootFinding
 using Base.Test
 
+const Interval = IntervalArithmetic.Interval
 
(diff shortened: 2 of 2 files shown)


dpsanders:
Thanks, LGTM. Not sure what's up with the `bisect` function.

lbenet:
The problem is not `bisect` itself, but using for that test the `a..b` notation, which uses `prevfloat(a)` if `a` is not an integer and something similar for `b` with `nextfloat(b)`. 

That's the reason we get:
```
julia> using IntervalArithmetic

julia> 0.5..0.5
[0.499999, 0.500001]

julia> inf(ans), sup(ans)
(0.49999999999999994, 0.5000000000000001)
```

`bisect` (correctly) returns a tuple of intervals constructed with `Interval`.

I just submitted a new commit which fixes the test.

lbenet:
Tests pass now in Julia 0.6, but fail in Julia 0.7. 

Any ideas?

dpsanders:
No, sorry, it's not clear what the problem is.

Can we use FemtoCleaner to get rid of all the deprecations?

lbenet:
We should certainly use it. Perhaps #14 and #17 should be merged as well.

lbenet:
Restarted failing travis jobs on nightly. Let's see if https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/93 is related...

lbenet:
Closing, since it is currently [reexported](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/96)

PR #38, opened 2018-03-08
Add tolerance check for `newton_refine`

### LINKED ISSUE #36: Tolerance in Newton is not respected
The tolerance setting doesn't seem to be respected by Newton.

eeshan9815:
Fixes #36 
@dpsanders Is this the expected behaviour? Can I have some test cases to try?
tolerance-newton


dpsanders:
`newton_refine` is no longer used.

The tolerance issue is in the `refine` function in `branch_and_prune.jl`.

PR #39, opened 2018-03-10
Interval Newton Method

eeshan9815:
Follows the Interval Newton Method as given in the book by Eldon Hansen, G. William Walster: Global Optimization Using Interval Analysis

### DIFF
--- a/src/newton1d.jl
+++ b/src/newton1d.jl
@@ -1,0 +1,30 @@
+function newton1d{T}(f::Function, f_prime::Function, x::Interval{T};
+                    epsX=eps(T), epsf=eps(T), debug=false)
+
+    L = Interval[]
+    R = Root[]
+    z = zero(x.lo)
+    push!(L, x)
+    while !isempty(L)
+        X = pop!(L)
+        mid_x = mid(X)
+        if !(z in f_prime(X))
+            # O ∉ f'(X)
+            while true
+                mid_x = mid(X)
+                N = mid_x - (f(Interval(mid_x))/f_prime(X))
+                X = X ∩ N
+                if isempty(X)
+                    break
+                elseif z in f(Interval(prevfloat(mid_x), nextfloat(mid_x)))
+                    push!(R, Root(X, :unique))
+                    break
+                end
+            end
+        else
+            # 0 ∈ f'(X)
+        end
+
+    end
+    return R
+end
--- a/src/newton1d.jl
+++ b/src/newton1d.jl
@@ -1,11 +1,11 @@
 """`newton1d` performs the interval Newton method on the given function `f`
 with its derivative `f′` and initial interval `x`.
-Optional keyword arguments give the tolerances `epsX` and `epsf`.
-`epsX` is the tolerance on the relative error whereas `epsf` is the tolerance on |f(X)|,
+Optional keyword arguments give the tolerances `reltol` and `abstol`.
+`reltol` is the tolerance on the relative error whereas `abstol` is the tolerance on |f(X)|,
 and a `debug` boolean argument that prints out diagnostic information."""
 
 function newton1d{T}(f::Function, f′::Function, x::Interval{T};
-                    epsX=eps(T), epsf=eps(T), debug=false)
+                    reltol=eps(T), abstol=eps(T), debug=false)
 
     L = Interval{T}[]
 
@@ -51,8 +51,8 @@ function newton1d{T}(f::Function, f′::Function, x::Interval{T};
                     expansion_pt = X.hi
 
                 else
-                    x1 = 0.25 * (3 * X.lo + X.hi)
-                    x2 = 0.25 * (X.lo + 3 * X.hi)
+                    x1 = mid(Interval(X.lo, mid(X)))
+                    x2 = mid(Interval(mid(X), X.hi))
                     if 0 ∉ f(Interval(x1)) || 0 ∉ f(Interval(x2))
                         push!(L, Interval(X.lo, m))
                         push!(L, Interval(m, X.hi))
@@ -67,7 +67,7 @@ function newton1d{T}(f::Function, f′::Function, x::Interval{T};
             else
                 # 0 ∉ fⁱ(x)
 
-                if (diam(X)/mag(X)) < epsX && diam(f(X)) < epsf
+                if (diam(X)/mag(X)) < reltol && diam(f(X)) < abstol
                     push!(R, Root(X, :unknown))
                     continue
                 end
@@ -120,8 +120,8 @@ end
 
 
 """`newton1d` performs the interval Newton method on the given function `f` and initial interval `x`.
-Optional keyword arguments give the tolerances `epsX` and `epsf`.
-`epsX` is the tolerance on the relative error whereas `epsf` is the tolerance on |f(X)|,
+Optional keyword arguments give the tolerances `reltol` and `abstol`.
+`reltol` is the tolerance on the relative error whereas `abstol` is the tolerance on |f(X)|,
 and a `debug` boolean argument that prints out diagnostic information."""
 
 newton1d{T}(f::Function, x::Interval{T};  args...) =
--- a/src/newton1d.jl
+++ b/src/newton1d.jl
@@ -16,15 +16,59 @@ function newton1d{T}(f::Function, f_prime::Function, x::Interval{T};
                 X = X ∩ N
                 if isempty(X)
                     break
-                elseif z in f(Interval(prevfloat(mid_x), nextfloat(mid_x)))
+                elseif z in f(Interval(prevfloat(mid_x), nextfloat(mid_x))) # What if the root is one of the end points of the interval. That way, this condition will never satisfy even thoughthe interval becomes very small. Explicit check for that?
                     push!(R, Root(X, :unique))
                     break
                 end
             end
         else
             # 0 ∈ f'(X)
-        end
+            pointOfExpansion = Inf
+            if z in f(Interval(mid(X)))
+                # 0 ∈ fⁱ(x)
+                # Step 7
 
+                if !(z in f(Interval(X.lo)))
+                    pointOfExpansion = X.lo
+                elseif !(z in f(Interval(X.hi)))
+                    pointOfExpansion = X.hi
+                else
+                    x1 = 0.25 * (3 * X.lo + X.hi)
+                    x2 = 0.25 * (X.lo + 3 * X.hi)
+                    if !(z in f(Interval(x1))) || !(z in f(Interval(x2)))
+                        push!(L, Interval(X.lo, mid_x))
+                        push!(L, Interval(mid_x, X.hi))
+                        break
+                    else
+                        push!(R, Root(X, :unique))
+                        break
+                    end
+                end
+            else
+                # 0 ∉ fⁱ(x)
+                if (diam(X)/mag(X)) < epsX && diam(f(X)) < epsf
+                    push!(R, Root(X, :unknown))
+                    break
+                end
+            end
+            # Step 8
+            if isinf(pointOfExpansion)
+                pointOfExpansion = mid(X)
+            end
+            initial_width = diam(X)
+            N = pointOfExpansion - (f(Interval(pointOfExpansion))/f_prime(X))
+            X = X ∩ N
+            mid_x = mid(X)
+            if isempty(X)
+                break
+            end
+            # How can a Newton step in (9.2.2) return two intervals?
+            if diam(X) > initial_width/2
+                push!(L, Interval(mid_x, X.hi))
+                X = Interval(X.lo, mid_x)
+            end
+            push!(L, X)
+        end
     end
     return R
 end
--- a/src/newton1d.jl
+++ b/src/newton1d.jl
@@ -1,4 +1,4 @@
-doc"""`newton1d` performs the interval Newton method on the given function `f`
+"""`newton1d` performs the interval Newton method on the given function `f`
 with its derivative `f′` and initial interval `x`.
 Optional keyword arguments give the tolerances `epsX` and `epsf`.
 `epsX` is the tolerance on the relative error whereas `epsf` is the tolerance on |f(X)|,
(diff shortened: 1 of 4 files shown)


eeshan9815:
@dpsanders 
[WIP] Currently implemented for case 0 ∉ f'(X)

eeshan9815:
@dpsanders

dpsanders:
Thanks! Could you please add some tests.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+        mid_x = mid(X)
+        if !(z in f_prime(X))
+            # O ∉ f'(X)
+            while true
+                mid_x = mid(X)
+                N = mid_x - (f(Interval(mid_x))/f_prime(X))
+                X = X ∩ N
+                if isempty(X)
+                    break
+                elseif z in f(Interval(prevfloat(mid_x), nextfloat(mid_x))) # What if the root is one of the end points of the interval. That way, this condition will never satisfy even thoughthe interval becomes very small. Explicit check for that?
```
eeshan9815: @dpsanders 
What if the root is one of the end points of the interval. That way, this condition will never satisfy even though the interval becomes very small. Should there be an explicit check for that?

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+                pointOfExpansion = mid(X)
+            end
+            initial_width = diam(X)
+            N = pointOfExpansion - (f(Interval(pointOfExpansion))/f_prime(X))
+            X = X ∩ N
+            mid_x = mid(X)
+            if isempty(X)
+                break
+            end
+            # How can a Newton step in (9.2.2) return two intervals?
```
eeshan9815: How can a Newton step in Section (9.2.2) return two intervals as given in the algorithm?

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
+function newton1d{T}(f::Function, f_prime::Function, x::Interval{T};
+                    epsX=eps(T), epsf=eps(T), debug=false)
+
+    L = Interval[]
```
dpsanders: Should be concretely typed.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+    R = Root[]
```
dpsanders: Also should be concretely typed.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+    z = zero(x.lo)
+    push!(L, x)
+    while !isempty(L)
```
dpsanders: Please add blank lines to separate pieces of code, and/or split up big functions into sub-functions.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+        X = pop!(L)
+        mid_x = mid(X)
+        if !(z in f_prime(X))
```
dpsanders: `\notin` gives `∉`

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
+function newton1d{T}(f::Function, f_prime::Function, x::Interval{T};
```
dpsanders: You can, if you want, use `f′`. You get the prime as `\prime<TAB>`

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+                    epsX=eps(T), epsf=eps(T), debug=false)
```
dpsanders: What are `epxX` and `epsf`?

src/newton1d.jl
dpsanders: Add a docstring for the function.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+
+    L = Interval[]
+    R = Root[]
+    z = zero(x.lo)
+    push!(L, x)
+    while !isempty(L)
+        X = pop!(L)
+        mid_x = mid(X)
```
dpsanders: Can just use `m` instead of `mid_x`.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+        if !(z in f_prime(X))
+            # O ∉ f'(X)
+            while true
+                mid_x = mid(X)
+                N = mid_x - (f(Interval(mid_x))/f_prime(X))
```
dpsanders: Add spaces around `/`

src/newton1d.jl
dpsanders: I think you can just use `0` instead of `z`.

dpsanders: I don't understand what this line is for.

dpsanders: I'm not sure what you mean?

src/newton1d.jl
dpsanders: What happens when `f'(X)` contains 0?

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+                if isempty(X)
+                    break
+                elseif z in f(Interval(prevfloat(mid_x), nextfloat(mid_x))) # What if the root is one of the end points of the interval. That way, this condition will never satisfy even thoughthe interval becomes very small. Explicit check for that?
+                    push!(R, Root(X, :unique))
+                    break
+                end
+            end
+        else
+            # 0 ∈ f'(X)
+            pointOfExpansion = Inf
```
dpsanders: Prefer using underscores (or no underscores) to camel case:

`point_of_expansion` or better `expansion_pt`, and add a comment explaining what this is.

src/newton1d.jl
```diff
@@ -0,0 +1,74 @@
...
+            if z in f(Interval(mid(X)))
+                # 0 ∈ fⁱ(x)
+                # Step 7
+
+                if !(z in f(Interval(X.lo)))
+                    pointOfExpansion = X.lo
+                elseif !(z in f(Interval(X.hi)))
+                    pointOfExpansion = X.hi
+                else
+                    x1 = 0.25 * (3 * X.lo + X.hi)
```
dpsanders: It's unclear what this magic expression does.

src/newton1d.jl
dpsanders: There is a reference in the text to it using "extended interval arithmetic" from chapter 4 in this case.

eeshan9815:
Thanks a lot for all the code reviews! Really appreciated.
I have included all the major steps of the algorithm from Chapter 9, with some minor sub-steps missing due to ambiguity. I will include all of them, and make all the changes according to the reviews.
Should I copy existing test cases and make them run for `newton1d` or make my own?

dpsanders:
You can copy the existing tests and add new ones from the books and that you invent.

src/newton1d.jl
eeshan9815: I'm not sure what you mean. I have handled the case for `0 ∈ f'(x)` separately.

Okay, I'll incorporate that.

src/newton1d.jl
```diff
@@ -0,0 +1,99 @@
+doc"""`newton1d` performs the interval Newton method on the given function `f`
```
dpsanders: I believe that `doc` is not necessary here. 
It is/was necessary when using `$` for LaTeX strings, but this can be avoided using double ``

src/newton1d.jl
```diff
@@ -0,0 +1,99 @@
...
+with its derivative `f′` and initial interval `x`.
+Optional keyword arguments give the tolerances `epsX` and `epsf`.
+`epsX` is the tolerance on the relative error whereas `epsf` is the tolerance on |f(X)|,
+and a `debug` boolean argument that prints out diagnostic information."""
+
+function newton1d{T}(f::Function, f′::Function, x::Interval{T};
+                    epsX=eps(T), epsf=eps(T), debug=false)
+
+    L = Array{Interval, 1}()
```
dpsanders: This is still not concretely typed, since `Interval` is an abstract type.

```
Interval{T}[]
```

src/newton1d.jl
```diff
@@ -0,0 +1,99 @@
...
+
+    R = Array{Root, 1}()
```
dpsanders: Same.

src/newton1d.jl
```diff
@@ -0,0 +1,99 @@
...
+
+    push!(L, x)
+
+    while !isempty(L)
+        X = pop!(L)
+        m = mid(X)
+
+        if 0 ∉ f′(X)
+            # O ∉ f'(X)
```
dpsanders: Remove this line.

dpsanders:
I don't see here the key part of the 1D Newton method, which you were asking about: what happens when 0 ∈ f′(X).

In this case, when you divide by f′(X), what happens, according to the definition of the division as a set? (I think chapter 4 of Hansen is probably overkill for this; look at Tucker).

src/newton1d.jl
eeshan9815: Ohh okay, thank you!

eeshan9815:
@dpsanders I have added the extended interval division case for the newton step and added the tests. Is there anything more to be done?

test/newton1d.jl
```diff
@@ -0,0 +1,41 @@
...
+big_pi = @interval(pi)
+# Using precision "only" 256 leads to overestimation of the true roots for `cos`
+# i.e the Newton method gives more accurate results!
+
+half_pi = big_pi / 2
+three_halves_pi = 3*big_pi/2
+
+@testset "Testing newton1d" begin
+
+    rts =  newton1d(sin, cos, -5..5)
```
dpsanders: You should repeat each test where you use AD to calculate the derivative.
You can do this using a loop.

test/newton1d.jl
```diff
@@ -0,0 +1,41 @@
...
+    @test length(rts) == 3
+    @test (-π.. -π) == rts[1].interval && :unique == rts[1].status
+    @test (0.. 0) == rts[2].interval && :unique == rts[2].status
```
dpsanders: You need the space in `a..b` only when `b` starts with a `-` sign.

test/newton1d.jl
```diff
@@ -0,0 +1,41 @@
...
+    f′(x) = 2*x*e^(x^2) + sin(x)
+    rts = newton1d(f, f′, -∞..∞)
+    @test length(rts) == 1
+    @test (0.. 0) == rts[1].interval && :unique == rts[1].status
+
+    f(x) = x^4 - 10x^3 + 35x^2 - 50x + 24
+    f′(x) = 4x^3 - 30x^2 + 70x - 50
+    rts = newton1d(f, f′, -10..10)
+    @test length(rts) == 4
+    @test 1 in rts[1].interval && :unique == rts[1].status
```
dpsanders: You should find a way to do this with a loop instead of by hand.
Have a look at the other tests where various loops are done.

src/newton1d.jl
```diff
@@ -0,0 +1,128 @@
+"""`newton1d` performs the interval Newton method on the given function `f`
+with its derivative `f′` and initial interval `x`.
+Optional keyword arguments give the tolerances `epsX` and `epsf`.
+`epsX` is the tolerance on the relative error whereas `epsf` is the tolerance on |f(X)|,
+and a `debug` boolean argument that prints out diagnostic information."""
+
+function newton1d{T}(f::Function, f′::Function, x::Interval{T};
+                    epsX=eps(T), epsf=eps(T), debug=false)
```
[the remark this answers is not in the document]
eeshan9815: According to the book by Hansen, 
"Note that we can set the tolerances ε X and ε f equal to zero. In this
case, a final condition before an interval is accepted as a final bound is that
0 ∈ f(x)"
That's why I set the default values as eps(T)

src/newton1d.jl
```diff
@@ -0,0 +1,128 @@
...
+                # Step 7
+
+                if 0 ∉ f(Interval(X.lo))
+                    expansion_pt = X.lo
+
+                elseif 0 ∉ f(Interval(X.hi))
+                    expansion_pt = X.hi
+
+                else
+                    x1 = 0.25 * (3 * X.lo + X.hi)
```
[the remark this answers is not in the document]
eeshan9815: Okay, twice, right?

src/newton1d.jl
```diff
@@ -0,0 +1,128 @@
...
+            end
+
+            initial_width = diam(X)
+
+            a = f(Interval(expansion_pt))
+            b = f′(X)
+
+            if 0 < b.hi && 0 > b.lo && 0 ∉ a
+                if a.hi < 0
+                    push!(L, X ∩ (expansion_pt - Interval(-Inf, a.hi / b.hi)))
```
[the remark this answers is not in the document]
eeshan9815: I thought about doing that, but it'll also need to return the number of intervals in the output along with the list of intervals. Should I incorporate it?

src/newton1d.jl
```diff
@@ -0,0 +1,128 @@
...
+        if 0 ∉ f′(X)
+            while true
+                m = mid(X)
+                N = m - (f(Interval(m)) / f′(X))
+                X = X ∩ N
+
+                if isempty(X)
+                    break
+
+                elseif 0 ∈ f(Interval(prevfloat(m), nextfloat(m)))
```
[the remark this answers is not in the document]
eeshan9815: It's the step 4 of the algorithm.
"Iterate the Newton method until either the result is empty or else
0 ∈ f (m(X))."
This part corresponds to 0 ∈ f (m(X)).

test/newton1d.jl
eeshan9815: The test case fails for f(x) = e^(x^2) - cos(x) as I don't think AD is able to calculate the derivative for that function, should I exclude that from AD?

test/newton1d.jl
```diff
@@ -0,0 +1,53 @@
...
+# Using precision "only" 256 leads to overestimation of the true roots for `cos`
+# i.e the Newton method gives more accurate results!
+
+half_pi = big_pi / 2
+three_halves_pi = 3*big_pi/2
+
+@testset "Testing newton1d" begin
+
+    f(x) = e^(x^2) - cos(x)
+    f′(x) = 2*x*e^(x^2) + sin(x)
```
dpsanders: Write these as exp(x^2) and then they will work. In general you should use exp instead of e

dpsanders:
Looks good, thanks a lot for the contribution!

PR #43, opened 2018-03-12
Bisect by default not exactly in centre

dpsanders:


dpsanders:
This is to avoid problems when bisecting a symmetric interval like `(-10..10)` exactly in the centre.
If there is a root exactly at 0, the method cannot find it since it is on the boundary of an interval.

dpsanders:
Superseded by #55.

PR #51, opened 2018-04-12
Explicitely give derivative to Newton contractor.

### LINKED ISSUE #45: Allow passing explicit derivative to Newton

Kolaru:
Fixes #45 

The tests mainly consist of comparing the default `Newton` method which uses `ForwardDiff` and passing the derivative/jacobian generated by `ForwardDiff` with the new syntax

    roots(f, X, Newton(f_prime))

### DIFF
--- a/src/branch_and_prune.jl
+++ b/src/branch_and_prune.jl
@@ -20,11 +20,10 @@ abstract type Contractor{F} end
 export Bisection, Newton
 
 struct Bisection{F} <: Contractor{F}
-    dimension::Int
     f::F
 end
 
-Bisection{n}(::Type{Val{n}}, f) = Bisection(n, f)
+Bisection{n}(::Type{Val{n}}, f) = Bisection(f)
 
 function (contractor::Bisection)(X)
     image = contractor.f(X)
@@ -38,7 +37,6 @@ end
 
 
 struct Newton{F,FP,O} <: Contractor{F}
-    dimension::Int
     f::F
     fp::FP
     op::O
@@ -46,16 +44,16 @@ end
 
 function Newton(::Type{Val{1}}, f::Function)
     f_prime = x -> ForwardDiff.derivative(f, x)
-    Newton(1, f, f_prime, N)
+    Newton(f, f_prime, N)
 end
 
 function Newton{n}(::Type{Val{n}}, f::Function)
     f_prime = x -> ForwardDiff.jacobian(f, x)
-    Newton(n, f, f_prime, N)
+    Newton(f, f_prime, N)
 end
 
 function Newton(f_prime::Function)
-    NewtonConstructor(f_prime)
+    return NewtonConstructor(f_prime)
 end
 
 function (C::Newton)(X)
@@ -79,52 +77,26 @@ function (C::Newton)(X)
     return :unknown, NX
 end
 
-"""
-    NewtonConstructor{FP}
 
-    Allow to store the derivative of a function before the creation of the
-    actual `Newton` contractor.
-"""
 struct NewtonConstructor{FP}
-    fp::FP
-end
-
-function (NC::NewtonConstructor){n}(::Type{Val{n}}, f)
-    Newton(n, f, NC.fp, N)
+    f_prime::FP
 end
 
 
 """
-    branch_and_prune(f, X, contractor, tol=1e-3)
+    branch_and_prune(X, contract, tol=1e-3)
 
-Generic branch and prune routine for finding isolated roots of a function
-`f:R^n → R^n` in a box.
+Generic branch and prune routine for finding isolated roots using the `contract`
+function as the contractor.
 
 Inputs:
-- `f`: function whose roots will be found
 - `X`: `Interval` or `IntervalBox`
-- `contractor`: function that, when applied to the function `f`, determines
-    the status of a given box `X`. It returns the new box and a symbol indicating
-    the status. Current possible values are `Bisection`, `Newton` and
-    `Newton(f_prime)` where `f_prime` is the derivative or jacobian of `f`.
+- `contractor`: function that determines the status of a given box `X`. It
+    returns the new box and a symbol indicating the status. Current possible
+    values are of type `Bisection` or `Newton`
 
 """
-function branch_and_prune(f, X, contractor, tol=1e-3)
-
-    input_dim = length(X)
-    output_dim = length(X)
-
-    # @show input_dim
-    # @show output_dim
-
-    # if !(input_dim == output_dim)
-    #     throw(ArgumentError("Input dimension ($input_dim) and output dimension ($output_dim) must be the same."))
-    # end
-
-    contract = contractor(Val{input_dim}, f)
-
-    # main algorithm:
-
+function branch_and_prune(X, contract, tol=1e-3)
     working = [X]
     outputs = Root{typeof(X)}[]
 
@@ -151,23 +123,11 @@ function branch_and_prune(f, X, contractor, tol=1e-3)
 
             push!(working, X1, X2)
         end
-
     end
 
     return outputs
 end
 
-branch_and_prune(f, X::Root, contractor, tol=1e-3) =
-    branch_and_prune(f, X.interval, contractor, tol)
-
-
-function branch_and_prune(f, V::Vector{Root{T}}, contractor, tol=1e-3) where {T}
-    reduce(append!, Root{T}[], [branch_and_prune(f, X.interval, contractor, tol) for X in V])
-end
-
-function branch_and_prune(f, V::Vector{T}, contractor, tol=1e-3) where {T}
-    reduce(append!, Root{T}[], [branch_and_prune(f, X, contractor, tol) for X in V])
-end
 
 export recursively_branch_and_prune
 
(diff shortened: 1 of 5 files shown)


src/branch_and_prune.jl
```diff
@@ -5,11 +5,94 @@ export branch_and_prune, Bisection, Newton
...
+    Newton(1, f, f_prime, N)
+end
+
+function Newton{n}(::Type{Val{n}}, f::Function)
+    f_prime = x -> ForwardDiff.jacobian(f, x)
+    Newton(n, f, f_prime, N)
+end
+
+function Newton(f_prime::Function)
+    NewtonConstructor(f_prime)
```
dpsanders: I don't see the need for a separate `NewtonConstructor` object here?

dpsanders:
Many thanks for the contribution!

src/branch_and_prune.jl
Kolaru: It allows to store the derivative which is needed to handle complex function the same way they are in the default case i.e. to turn the complex derivative into a real jacobian.

However it is true that `Newton(f_prime::Function)` could simply return `f_prime`, and we could dispatch on the fact it is a function rather than some custom type. But then if we implement another method possibly using the derivative (I believe Krawcyzk method allows it but I still have to check it in detail), we lose the information about what method it is.

Kolaru: I definitively have to improve my "code explaining skills" :D

The short and hopefully clearer answer about the constructor is that it is the handiest solution I found to be able to
1 Track the expected contractor (in this case `Newton`).
2 Keep the provided derivative easily accessible for manipulation until the contractor is created.

The second part being only useful for complex functions.

Anyway, I just figured that this can be more elegantly done by moving the instancing of the contractor out of `branch_and_prune` to the `roots` function. So that `branch_and_prune` take a contractor instance rather than a contractor type (or constructor) as third parameter.

This will also resolve any concern about type stability I think.

And finally, no, I did not do any benchmark.

Kolaru:
(I apparently messed up the message of the last commit, git sometimes eludes me)

Here is a short summary of the changes in the new commit:
1. `branch_and_prune` is no longer overloaded and now always has the signature `branch_and_prune(X, contract, tol=1e-3)`. The function to be applied is no longer passed as an argument as it is not explicitly needed (it is accessed through the separator).
2. As a consequence, `roots` is now fully responsible to correctly create the separator.
3. This allows to no longer store the dimension in the separator as it is only needed to create the derivative/jacobian when it is not explicitly given. In turn it solve the question of making the dimension a type parameter.
4. However despite my hopes (and promises), `Newton(f_prime::Function)` still returns a `NewtonConstructor`. The reason for that is that having the two syntaxes and `roots(f, X, Newton)` and `roots(f, X, Newton(f_prime))` has the following consequence: either `Newton` and `Newton(f_prime)` return the same type or not.

In the first case, whether an explicit derivative have been passed must be somehow checked in `roots` and if it is the case the function `f` must be added to the separator. This makes the code both messy and kind of convoluted.

In the second case the fact that `Newton` and `Newton(f_prime)` return different types can be used to distinguish the two syntaxes in `roots`. This is the choice to which I have sticked for now. However despite the fact that `Newton` can return multiple types, the code is still type stable: for a given signature of the `Newton`, the function always has the same return type, so it can be inferred.

Hope this explanation helps.

dpsanders:
Sorry, I wasn't suggesting your code was bad, but rather that my original code needed a refactor, but it sounds like you've done it already -- thanks! I'll take a careful look.

But I see the following simple benchmarks:

Current master:
```
 @time roots(x->x^2 - 2, -10..10)
  0.043622 seconds (21.65 k allocations: 1.096 MiB)
2-element Array{IntervalRootFinding.Root{IntervalArithmetic.Interval{Float64}},1}:
 Root([1.41421, 1.41422], :unique)
 Root([-1.41422, -1.41421], :unique)
```
This PR:
```
 0.058387 seconds (27.36 k allocations: 1.389 MiB)
```
which is a 33% slowdown.

src/branch_and_prune.jl
```diff
@@ -5,42 +5,98 @@ export branch_and_prune, Bisection, Newton
...
 
-    # main algorithm:
+Inputs:
+- `X`: `Interval` or `IntervalBox`
+- `contractor`: function that determines the status of a given box `X`. It
+    returns the new box and a symbol indicating the status. Current possible
+    values are of type `Bisection` or `Newton`
 
+"""
+function branch_and_prune(X, contract, tol=1e-3)
```
dpsanders: `contract` -> `contractor`

src/branch_and_prune.jl
```diff
@@ -5,42 +5,98 @@ export branch_and_prune, Bisection, Newton
...
+struct NewtonConstructor{FP}
+    f_prime::FP
+end
+
 
-    # if !(input_dim == output_dim)
-    #     throw(ArgumentError("Input dimension ($input_dim) and output dimension ($output_dim) must be the same."))
-    # end
+"""
+    branch_and_prune(X, contract, tol=1e-3)
```
dpsanders: Can this function just be renamed `roots` now?

src/branch_and_prune.jl
```diff
@@ -221,9 +181,45 @@ Inputs:
...
 - `contractor`: function that, when applied to the function `f`, determines
     the status of a given box `X`. It returns the new box and a symbol indicating
-    the status. Current possible values are `Bisection` and `Newton`.
+    the status. Current possible values are `Bisection`, `Newton` and
+    `Newton(f_prime)` where `f_prime` is the derivative or jacobian of `f`.
 
 """
-roots{C<:Contractor}(f, X, contractor::Type{C}, tol::Float64=1e-3) = branch_and_prune(f, X, contractor, tol)
+function roots{T, C<:Contractor}(f, X::IntervalLike{T}, contractor::Type{C}, tol::Float64=1e-3)
+    dim = length(X)
```
dpsanders: Can the length of `X` just be a type parameter of the `roots` function?

```
+function roots(f, X::IntervalBox{T,N}, contractor::Type{C}, tol::Float64=1e-3) where {T, N, C<:Contractor}
```
and just wrap an Interval in a 1D IntervalBox? Or just provide a separate function altogether for the 1D case.

dpsanders: (Please use the future-proof `where` syntax.)

src/branch_and_prune.jl
```diff
@@ -221,9 +181,45 @@ Inputs:
...
+    contract = contractor(Val{dim}, f)
+    branch_and_prune(X, contract, tol)
+end
+
+function roots{T}(f, X::IntervalLike{T}, nc::NewtonConstructor, tol::Float64=1e-3)
+    dim = length(X)
+    contract = Newton(f, nc.f_prime, N)
```
dpsanders: `contractor`

dpsanders: I think we should rename the Newton operator `N` as `NN` to avoid confusion.

dpsanders: Or even the script 𝒩 (`\mscrN`)

src/branch_and_prune.jl
```diff
@@ -5,42 +5,98 @@ export branch_and_prune, Bisection, Newton
...
+end
+
+
+struct Newton{F,FP,O} <: Contractor{F}
+    f::F
+    fp::FP
+    op::O
+end
+
+function Newton(::Type{Val{1}}, f::Function)
```
dpsanders: I know this was in my original code, but it seems nasty to have the `Val{1}` stuff.
I think this was basically to get around the fact that you can't extract the dimensions from the function `f`.

Maybe we should have some kind of wrapper for the function that contains the information on its dimensions.

dpsanders:
This is looking quite a lot cleaner than the original code, thanks!

I still haven't got my head around the need for `NewtonConstructor`.

Kolaru:
The performance difference is a deep mystery for me. If I run it using `BenchmarkTools` I get for current master

      @benchmark roots(x -> x^2 - 2, -10..10)
      BenchmarkTools.Trial:
        memory estimate:  163.00 KiB
        allocs estimate:  4535
        --------------
        minimum time:     442.071 μs (0.00% GC)
        median time:      513.917 μs (0.00% GC)
        mean time:        592.733 μs (11.30% GC)
        maximum time:     11.005 ms (93.67% GC)
        --------------
        samples:          8410
        evals/sample:     1

and for this PR:

    @benchmark roots(x -> x^2 - 2, -10..10)
    BenchmarkTools.Trial:
      memory estimate:  160.39 KiB
      allocs estimate:  4444
      --------------
      minimum time:     424.843 μs (0.00% GC)
      median time:      510.619 μs (0.00% GC)
      mean time:        578.947 μs (11.10% GC)
      maximum time:     11.543 ms (93.58% GC)
      --------------
      samples:          8606
      evals/sample:     1

It look like there is a difference only if the function is run once. For example in current master we have:

    @time for i in 1:100 roots(x -> x^2 -2, -10..10) end
      0.096897 seconds (470.96 k allocations: 16.853 MiB, 10.24% gc time)

and for this PR:

    @time for i in 1:100 roots(x -> x^2 -2, -10..10) end
      0.105922 seconds (467.67 k allocations: 16.891 MiB, 9.34% gc time)

I have no idea where it could come from (and profiling will probably not help either).

And the need for a `NewtonConstructor` is indeed not obvious, but I think it is the best (if not only) satisfying alternative for the current syntax (I would not have imagined it either if I hadn't tried to implement that ^^).

src/branch_and_prune.jl
Kolaru: It could as no other `roots` function conflict with it. However from a design point of view separating the function actually containing the algorithm from the setup function (in our case `roots`) may add readability.

On the other hand if it is named `roots` it will be exported, allowing for syntax like

    roots(x -> x^2 - 2, Newton(f, f_prime), -10..10

src/branch_and_prune.jl
Kolaru: If we decide that the `roots` function (rather than the `Newton` one) is responsible to provide derivative, even when none was explicitly given, the dimension is no longer useful as it is only used to choose between `ForwardDiff.derivative` and `ForwardDiff.jacobian`.

Kolaru: It could be done by specializing the `roots` function even more:

    function roots(f, X::Interval{T}, contractor::Newton, tol::Float64=1e-3) where {T}
        contract = contractor(f, x -> ForwardDiff.derivative(f, x)))
        branch_and_prune(X, contract, tol)
    end

    function roots(f, X::IntervalBox{T}, contractor::Newton, tol::Float64=1e-3) where {T}
        contract = contractor(f, x -> ForwardDiff.jacobian(f, x)))
        branch_and_prune(X, contract, tol)
    end

Kolaru:
I just fix a bit of notation and simplified the `Newton` as discussed. Now I thing it is ready to be merged :)

dpsanders:
Very nice, thanks a lot!

PR #57, opened 2018-04-26
Implement Krawczyk method

Kolaru:
- Implement new contractor type `Krawczyk` and Krawczyk operator `𝒦`
- Implement related tests
- Clean up and uniformized tests

### DIFF
--- a/src/contractors.jl
+++ b/src/contractors.jl
@@ -6,7 +6,7 @@
 """
 abstract type Contractor{F} end
 
-export Bisection, Newton
+export Bisection, Newton, Krawczyk
 
 # bisection
 struct Bisection{F} <: Contractor{F}
@@ -42,10 +42,7 @@ function (C::Newton)(X, tol)
     NX = 𝒩(C.f, C.f′, X) ∩ X
 
     isempty(NX) && return :empty, X
-
-    if isinf(X)
-        return :unknown, NX  # force bisection
-    end
+    isinf(X) && return :unknown, NX  # force bisection
 
     if NX ⪽ X  # isinterior; know there's a unique root inside
         NX =  refine(X -> 𝒩(C.f, C.f′, X), NX, tol)
@@ -72,13 +69,10 @@ function 𝒩{T}(f, f′, X::Interval{T})
 end
 
 
-IntervalArithmetic.mid(X::IntervalBox, α) = mid.(X, α)
-
 doc"""
 Multi-variable Newton operator.
 """
 function 𝒩(f::Function, jacobian::Function, X::IntervalBox)  # multidimensional Newton operator
-
     m = IntervalBox(Interval.(mid(X, where_bisect)))
     J = jacobian(SVector(X))
 
@@ -86,6 +80,58 @@ function 𝒩(f::Function, jacobian::Function, X::IntervalBox)  # multidimension
 end
 
 
+# Krawczyk
+struct Krawczyk{F, FP} <: Contractor{F}
+    f::F
+    f′::FP   # use \prime<TAB> for ′
+end
+
+function (C::Krawczyk)(X, tol)
+    # use Bisection contractor for this:
+    if !(contains_zero(C.f(X)))
+        return :empty, X
+    end
+
+    KX = 𝒦(C.f, C.f′, X) ∩ X
+
+    isempty(KX) && return :empty, X
+    isinf(X) && return :unknown, KX  # force bisection
+
+    if KX ⪽ X  # isinterior; know there's a unique root inside
+        KX =  refine(X -> 𝒦(C.f, C.f′, X), KX, tol)
+        return :unique, KX
+    end
+
+    return :unknown, KX
+end
+
+
+doc"""
+Single-variable Krawczyk operator
+"""
+function 𝒦(f, f′, X::Interval{T}) where {T}
+    m = Interval(mid(X, where_bisect))
+    Y = 1/f′(m)
+
+    m - Y*f(m) + (1 - Y*f′(X))*(X - m)
+end
+
+
+@generated function 𝒦(f, jacobian, X::IntervalBox{N, T}) where {N, T}
+    unit = eye(N)
+
+    ex = quote
+        m = mid(X, where_bisect)
+        J = jacobian(X)
+        Y = inv(jacobian(m))
+        m = IntervalBox(Interval.(m))
+
+        IntervalBox(m - Y*f(m) + ($unit - Y*J)*(X - m))
+    end
+    ex
+end
+
+IntervalArithmetic.mid(X::IntervalBox, α) = mid.(X, α)
 
 """
 Generic refine operation for Krawczyk and Newton.
--- a/src/contractors.jl
+++ b/src/contractors.jl
@@ -1,14 +1,20 @@
-# contractors:
-"""
+export Bisection, Newton, Krawczyk
+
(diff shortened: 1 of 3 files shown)


src/contractors.jl
```diff
@@ -72,20 +69,69 @@ function 𝒩{T}(f, f′, X::Interval{T})
...
+@generated function 𝒦(f, jacobian, X::IntervalBox{N, T}) where {N, T}
+    unit = eye(N)
+
+    ex = quote
+        m = mid(X, where_bisect)
+        J = jacobian(X)
+        Y = inv(jacobian(m))
+        m = IntervalBox(Interval.(m))
+
+        IntervalBox(m - Y*f(m) + ($unit - Y*J)*(X - m))
```
dpsanders: You should just be able to use `I` here and get rid of the `@generated`.

src/roots.jl
```diff
@@ -78,6 +78,7 @@ contains_zero(X::IntervalBox) = all(contains_zero.(X))
 
 
 IntervalLike{T} = Union{Interval{T}, IntervalBox{T}}
+DerivativeType = Union{Type{Newton}, Type{Krawczyk}}
```
dpsanders: Make call this `NewtonLike`?

dpsanders:
Looks good, thanks a lot!

src/contractors.jl
```diff
@@ -72,20 +69,69 @@ function 𝒩{T}(f, f′, X::Interval{T})
...
 end
 
 
+# Krawczyk
+struct Krawczyk{F, FP} <: Contractor{F}
+    f::F
+    f′::FP   # use \prime<TAB> for ′
+end
+
+function (C::Krawczyk)(X, tol)
```
dpsanders: This function is (almost?) exactly the same as the equivalent for Newton, which is why my idea had been to combine the two into a single type. It's not necessary to use a single type, but I think this function and the equivalent for Newton should be able to be combined.

dpsanders: E.g. define a function `operator(C::Krawczyk) = 𝒦`.
I *think* this should be efficient.

src/contractors.jl
```diff
@@ -72,20 +69,69 @@ function 𝒩{T}(f, f′, X::Interval{T})
...
+        J = jacobian(X)
+        Y = inv(jacobian(m))
+        m = IntervalBox(Interval.(m))
+
+        IntervalBox(m - Y*f(m) + ($unit - Y*J)*(X - m))
+    end
+    ex
+end
+
+IntervalArithmetic.mid(X::IntervalBox, α) = mid.(X, α)
```
dpsanders: I think we can get rid of this.

src/contractors.jl
```diff
@@ -72,20 +69,69 @@ function 𝒩{T}(f, f′, X::Interval{T})
...
+
+    m - Y*f(m) + (1 - Y*f′(X))*(X - m)
+end
+
+
+@generated function 𝒦(f, jacobian, X::IntervalBox{N, T}) where {N, T}
+    unit = eye(N)
+
+    ex = quote
+        m = mid(X, where_bisect)
```
dpsanders: I think we can get rid of `where_bisect`.

src/contractors.jl
```diff
@@ -72,20 +69,69 @@ function 𝒩{T}(f, f′, X::Interval{T})
...
 
 
-IntervalArithmetic.mid(X::IntervalBox, α) = mid.(X, α)
-
 doc"""
 Multi-variable Newton operator.
 """
 function 𝒩(f::Function, jacobian::Function, X::IntervalBox)  # multidimensional Newton operator
-
     m = IntervalBox(Interval.(mid(X, where_bisect)))
```
dpsanders: I think we can get rid of `where_bisect`, since calling `mid` without the second argument now does this automatically.

dpsanders: Ah yes, you're totally right, sorry!

dpsanders: Ouch that division is nasty...

dpsanders: Not sure what that should give.

test/roots.jl
```diff
@@ -66,32 +102,43 @@ end
     XX = IntervalBox(-5..6, 2)
     tol = 1e-5
 
+    # Newton
     rts = roots(gradf, XX, Newton, tol)
     @test length(rts) == 25
+    @test all_unique(rts)
     @test rts == roots(gradf, XX, Newton, tol; deriv = xx -> ForwardDiff.jacobian(gradf, xx))
+
+    # Krawczyk
```
dpsanders: This part and Newton above can be done with a `for` loop instead of duplicating the code.

src/contractors.jl
Kolaru: In fact we have

```julia
0/X == [0, 0]
```

for any `X` that contains 0.

A quick look at the IEEE standard did not help much (I may read that in full someday to really understand all the stuff about flavors and decorated intervals). Note that `extended_div` suffers the same problem for now

```julia
julia> extended_div(0..0, -1..1)
([0, 0], ∅)
```

but here the standard  definition (10.5.5 Two output division) indicates that it should return `([-∞, ∞], ∅)`.

dpsanders: Yes, I agree that `extended_div` is wrong here, according to the standard and comparing with the Octave interval arithmetic package; cc @eeshan9815 .

On the other hand, for standard division the situation is not so clear; the Octave package also returns `[0, 0]` in this case.

src/contractors.jl
```diff
@@ -23,38 +29,54 @@ function (contractor::Bisection)(X, tol)
...
 
-Base.isinf(X::IntervalBox) = any(isinf.(X))
+    Contraction operation for contractors using the first derivative of the
+    function. This contraction use a bisection scheme to refine the intervals
+    with `:unkown` status.
 
-function (C::Newton)(X, tol)
+    Currently `Newton` and `Krawczyk` contractors uses this.
+"""
+function newtonlike_contract(op, C, X, tol)
```
Kolaru: I am not sure this solution is ideal. However I feel like the most natural way of doing this would be to not have the contractors callable, but to have a method

```julia
contract(C::Contractor, X, tol)
```

and to use multiple dispatch/generated functions to handle everything correctly. Since it would be a deeper design change I wanted to discuss it first and it may be better in a separate PR.

dpsanders: That definitely sounds like an interesting idea. I'm not sure exactly what it would gain - I guess that's why a PR would be useful.

dpsanders:
Please don't include merge commits. I think you should be rebasing instead:

```
git checkout master
git pull upstream master

git checkout <your-branch>
git rebase master
```

Here, `upstream` is the remote that points to this repo.

dpsanders:
Ah and then `git push -f` to force push the changes to the current branch.

src/contractors.jl
dpsanders: I feel that it's more natural to call a contractor `C` as `C(X)` to reflect the mathematical structure. This is just as possible to dispatch on, no? Or what do you gain by introducing the `contract` function?

dpsanders:
Looks good to me, thanks a lot!

src/contractors.jl
dpsanders: Thanks, yeah I was trying to do what you're suggesting and ran into the same frustrating problem :/ 
Let's leave it like this for the moment.

Issue #22, opened 2017-06-27
Use queue instead of stack 


dpsanders:

eeshan9815:
@dpsanders Can you please point me to the place in the codebase this needs to be changed?

dpsanders:
This refers to the data structure used to store the boxes as they are
processed, and in particular the order in which they are processed.

Currently they are placed onto the vector with push! and removed with pop!
(?) which gives a stack (last in first out).

This comment was along the lines of using a queue instead (first in first
put).

dpsanders:
This gives breadth-first vs depth-first search of the tree of boxes that gets created.

gwater:
I think we should decide what the desired outcome of each strategy is.

**memory** In many cases depth-first strategy should be more memory efficient because after an interval is bisected its sub intervals are immediately pruned and quickly vanish from memory. A breadth-first search would bisect many intervals before eventually pruning their sub intervals, leading to potentially large fluctuations in queue size.

**time** Both strategies should have identical speed on a single process. However, depth-first will find Roots more periodically throughout the iteration, whereas breadth-first will find most roots close to the end of iteration.

**parallelism** However, a breadth-first strategy might be useful for a parallelized implementation. In that model work intervals *should* be bisected as early as possible to fill a task queue which can feed many worker processes. (Of course it would also look very nice in a visualization to watch intervals split, vanish and shrink in parallel. But aesthetics are probably irrelevant to most users.)

Did I miss something?

Kolaru:
I was thinking about #59 and I realized that in the case of an accumulation of many zeros (for example with `x -> sin(1/x)` near 0) the maximal number of intervals must be limited, but also bread-first or depth-first searches would result in very different outputs (still using `sin(1/x)` as an example):

- Depth-first search will find many negative zeros, but one of the output will be `Root(0..+Inf, :unkown)`.
- Breadth-first search will more or less find zeros uniformly.

Ideally both should be implemented, and it may be possible to even allow the user to provide a function to choose the next interval to be bisected (as a parameter of the `RootSearch` object for example, since it is low level tweeking).

Kolaru:
I am currently working on this (I let you know to avoid the work being done twice).

In fact I am trying to implement something more general, where the user can provide a data structure (currently we use `Vector`), a method to store the intervals (currently `push!`) and a method to get them (currently `pop!`).

gwater:
[...]

Issue #59, opened 2018-04-26
Limit the maximum number of intervals 


dpsanders:
Eg finding roots of sin on -Inf..Inf

Kolaru:
Covered by #178

PR #60, opened 2018-04-27
Implement Base.Iterator interface for branch and prune routine

gwater:
Hi, I found this package really useful so I thought I'd propose this feature:

It splits the `branch_and_prune()` routine up so it can be inspected at every iteration by implementing the [`Base.Iterator`](https://docs.julialang.org/en/stable/manual/interfaces/#man-interface-iteration-1) interface. A few other packages like DifferentialEquations.jl (for `Integrators`) and ProximalAlogrithms.jl use similar approaches. Iterators can be used to visualize the algorithms and probably other nifty things.

The changes have negligible impact on speed and memory. In fact, they *remove a minor type instability* from `branch_and_prune()` by making the type of `outputs` predictable. The existing API is unchanged, but a new type `RootSearch` is exported which implements the appropriate `Iterator` methods. (Introducing a type to indicate the search strategy might also be useful for #22.)

### DIFF
--- a/src/roots.jl
+++ b/src/roots.jl
@@ -1,7 +1,9 @@
 
 import IntervalArithmetic: diam, isinterior
+import Base: start, next, done, copy, eltype, iteratorsize
 
-export branch_and_prune, Bisection, Newton
+export branch_and_prune, Bisection, Newton, RootSearch
+export start, next, done, copy, step!, eltype, iteratorsize
 
 diam(x::Root) = diam(x.interval)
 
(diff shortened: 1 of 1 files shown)


dpsanders:
Hi, this is great, thanks a lot!

Could you please add an example and a test of how to use the iterator interface?

gwater:
Thanks, I added a simple use case to `examples/roots.jl`. However, I'm not sure what test case we need: The `Iterator` methods are almost completely covered by existing unit tests (through `branch_and_prune()`). We could use [InterfaceTesting.jl](https://github.com/oxinabox/InterfaceTesting.jl) to be really thorough but it seems they haven't caught up with julia 0.6 yet.

src/roots.jl
```diff
@@ -1,15 +1,77 @@
...
+    else # branch
+        X1, X2 = bisect(X)
+        push!(state.working, X1, X2)
+    end
+    return nothing
+end
+
+function next(iter::RootSearch, state::RootSearchState)
+    step!(state, iter.contractor, iter.tolerance)
+    return state, state
```
dpsanders: It seems strange to be returning `state` twice here?

dpsanders:
I like the idea of introducing a type for the search strategy.
I guess that's for a different PR? Or would you like to include it here?

examples/roots.jl
```diff
@@ -22,6 +22,15 @@ rts = roots(f, Xc, Bisection)
 rts = roots(f, rts, Newton)
 rts = roots(f, Xc)
 
+# track the number of working intervals during the iteration:
+f(x) = sin(x)
+contractor = Newton(f, x -> ForwardDiff.derivative(f, x))
+search = RootSearch(-10..10, contractor, 1e-3)
+for state in search
```
dpsanders: Very nice!

src/roots.jl
gwater: Yeah, it's a little awkward. The second return value must be `state` (dictated by the interface) and the first return value is what a `for i in iter` loop puts into `i`. Right now I think the most useful Iterator is one that allows inspection of the algorithms progress and therefore `i` should be the internal state.

One could imagine a *different* Iterator for which `next()` returns individual Roots as they are found but that will only be possible once the new [Nullable Iterator interface](https://github.com/JuliaLang/julia/issues/18823) is done. (Because we cannot decide if the iteration is done until we find (or dont find) the last root, so it is impossible to define a `done()` method.) 

(I should also point out that a lot of other Iterators use an Int as their state variable and modify the iterator object itself at every step which seems very inelegant and confusing to me.)

gwater:
I think the different search strategies should be dealt with in a seperate PR because they have a lot more impact on users than this mostly "under the hood" change. There will probably need to be some discussion about defaults, etc.

examples/roots.jl
gwater: Thanks :)

dpsanders:
I wonder if we could / should introduce the ability to call a callback function to facilitate interacting with the root search? In any case, I think this is good to merge.

Many thanks!

gwater:
Regarding, the callback function I'm of two minds: On one hand it would be very easy to implement and there would be no cost. On the other hand effectively *using* it would require detailed knowledge of the internals; in which case one could just do what I did in the example and the callback would be superfluous.

gwater:
It might be useful to expose some specific break conditions though:

* a maximum number of iterations
* a maximum number of intervals / memory limit (#59 )

It should be quite easy to include these in the revised `branch_and_prune()`.

dpsanders:
Thanks. I'm curious what your particular use case is?

gwater:
Sure, I've hinted at this a few times: I think it would be cool to visualize intervals splitting and vanishing/shrinking. I'm working on an eigenvalue problem for a delay differential equation and I'd like to give people an impression of the algorithm used to solve it.

PR #62, opened 2018-05-01
Add missing deriv

dpsanders:


lbenet:
Thanks! The sole comment i have is related to the default `tol`. In the fall-back method, `tol`'s default is 1.0e-15; other methods use 1.e-3. I think this is a good opportunity to fix a common default.

Shall I go ahead and merge (and leave the common `tol` default for another PR) or do you plan to make further changes?

dpsanders:
No more changes planned.
Yes, not sure what to do about the default tol.

lbenet:
Mwrging. We can set a common default for `tol` in another PR.

Thanks a lot!

PR #63, opened 2018-05-02
Bring back a method for 𝒩

lbenet:
This brings back a method for 𝒩 (which depends on the function `f` and two intervals) which was deleted when #55 was merged. It will probably not be needed once `find_roots` is deleted, but for the time being I have a case that requires it. (This method was untested previously; I haven't add any tests, but I could if it makes any sense.)


lbenet:
@dpsanders Tests pass, including appveyor.

dpsanders:
Great, thanks!

PR #65, opened 2018-05-07
Add deriv as another parameter of roots (instead of a keyword arg)

### LINKED ISSUE #64: Make derivative another parameter of roots?
E.g.

`roots(x->x^2 - 2, x->2x, X, tol)`

This is how it used to be. This removes the need for keyword arguments and improves type inference.

lbenet:
Fixes #64

### DIFF
--- a/src/roots.jl
+++ b/src/roots.jl
@@ -138,32 +138,41 @@ function roots(f, X::IntervalLike{T}, ::Type{Bisection}, tol::Float64=1e-3; deri
     branch_and_prune(X, Bisection(f), tol)
 end
 
-function roots(f, X::Interval{T}, C::NewtonLike, tol::Float64=1e-3; deriv = nothing) where {T}
+function roots(f, X::Interval{T}, C::NewtonLike, tol::Float64=1e-3) where {T}
 
-    if deriv == nothing
-        deriv = x -> ForwardDiff.derivative(f, x)
-    end
+    deriv = x -> ForwardDiff.derivative(f, x)
 
     branch_and_prune(X, C(f, deriv), tol)
 end
 
-function roots(f, X::IntervalBox{T}, C::NewtonLike, tol::Float64=1e-3; deriv = nothing) where {T}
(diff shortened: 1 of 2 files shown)


src/roots.jl
```diff
@@ -138,32 +138,41 @@ function roots(f, X::IntervalLike{T}, ::Type{Bisection}, tol::Float64=1e-3; deri
     branch_and_prune(X, Bisection(f), tol)
 end
 
-function roots(f, X::Interval{T}, C::NewtonLike, tol::Float64=1e-3; deriv = nothing) where {T}
+function roots(f, X::Interval{T}, C::NewtonLike, tol::Float64=1e-3) where {T}
```
dpsanders: You should probably just call the version of `roots` that has the explicit derivative argument here.

lbenet: Thanks for the comment; done, including some methods below.

lbenet:
In the new commit, I also deleted the dependency on `deriv` for `roots` using the `Bisection` contractor.

Kolaru:
I have no special, everything looks good to me.

lbenet:
@dpsanders Shall I merge this? I would also propose to tag a new patch version

dpsanders:
Yes please, go ahead and merge.
We should tag a minor version once all the current PRs are merged, since the API has changed (?)

lbenet:
My proposal is to tag a patch version only; I have some tests which require some merged PRs after v0.2.0.

lbenet:
@dpsanders I tried to tag a new patch version (v0.2.1), so I can keep on working on TaylorModels; yet, some dependencies had broken tests (see https://github.com/JuliaLang/METADATA.jl/pull/14657). Do you know how to try this again? I am not sure if I should delete the version and then tag a new one.

See also https://github.com/JuliaIntervals/ValidatedNumerics.jl/pull/284, which solves the issues for ValidatedNumerics.

dpsanders:
Yes, the easiest way is to delete the release and the tag and make the
release again.

PR #67, opened 2018-05-14
System of Linear Equations 

eeshan9815:


eeshan9815:
I manually benchmarked, and shared the results on #intervals in slack

dpsanders:
Please include at least the benchmarking script (using BenchmarkTools) in a new directory `perf`.

eeshan9815:
What should the script evaluate and return exactly? Also, the directory name should be `perf`?

eeshan9815:
cc @dpsanders

dpsanders:
That's the kind of thing I had in mind, thanks. You could instead collect the data using `@belapsed` (I think) and output a table using Dataframes or something like that.

eeshan9815:
Okay, I'll try it out!
So which version of the function should I retain? I don't think `MArray` improved the performance a lot.

dpsanders:
Try à version using IntervalBox and setindex.

dpsanders:
IntervalBox is immutable like an SVector. So modifying it actually requires making a new copy and changing the relevant entry using setindex (without !)

src/linear_eq.jl
```diff
@@ -87,3 +87,47 @@ function gauss_seidel_interval_static!{T, N}(x::MVector{N, Interval{T}}, A::MMat
...
+
+    return B*A, B*b
+
+end
+
+
+function gauss_seidel_interval_static1{T, N}(A::SMatrix{N, N, Interval{T}}, b::SVector{N, Interval{T}}; precondition=true, maxiter=100)
+
+    n = size(A, 1)
+    x = @MVector fill(-1e16..1e16, n)
```
dpsanders: `@SVector`?

dpsanders:
This needs tests.

src/linear_eq.jl
eeshan9815: I realized only `x` has to be modified, so I tried a version where `A` and `b` are `SVector`s, and that showed improvement.

eeshan9815:
> Isn't this just a different way of writing the same thing? (And probably slower, but maybe the code is clearer.)
Yes, it's just a vectorized way of writing the same thing, I later realized.

eeshan9815:
@dpsanders Travis and AppVeyor both use the latest tagged version and don't have `extended_div`, and hence are failing.

src/linear_eq.jl
dpsanders: Please try a version where `x` is an `SVector` and you use `setindex` to "modify" it (i.e. replace `x` with a new `SVector`).

dpsanders:
Actually the matrix version could be pretty efficient when using SMatrix and SVector everywhere. Have you tried that?

src/linear_eq.jl
eeshan9815: Done.
I think this version gives the best benchmarks.

eeshan9815:
Done.
I think this version gives the best benchmarks.

src/linear_eq.jl
```diff
@@ -131,6 +131,50 @@ function gauss_seidel_interval_static1!{T, N}(x::MVector{N, Interval{T}}, A::SMa
...
+"""
+Iteratively solves the system of interval linear
+equations and returns the solution set. Uses the
+Gauss-Seidel method (Hansen-Sengupta version) to solve the system.
+Keyword `precondition` to turn preconditioning off.
+Eldon Hansen and G. William Walster : Global Optimization Using Interval Analysis - Chapter 5 - Page 115
+"""
+function gauss_seidel_interval_static2!{T, N}(x::SVector{N, Interval{T}}, A::SMatrix{N, N, Interval{T}}, b::SVector{N, Interval{T}}; precondition=true, maxiter=100)
+
+    precondition && ((M, r) = preconditioner_static2(A, b))
```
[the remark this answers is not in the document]
eeshan9815: Fixed.

dpsanders:
So it seems just using normal arrays is fine?

eeshan9815:
> So it seems just using normal arrays is fine?

More or less, yes. Should I remove the other implementations and update the PR?

perf/linear_eq.jl
```diff
@@ -0,0 +1,40 @@
...
+    df = DataFrame()
+    df[:Method] = ["Array", "MArray", "SArray1", "SArray2", "Contractor"]
+    for n in 1:max
+        A, mA, sA = randMat(n)
+        b, mb, sb = randVec(n)
+        t1 = @belapsed gauss_seidel_interval($A, $b)
+        t2 = @belapsed gauss_seidel_interval_static($mA, $mb)
+        t3 = @belapsed gauss_seidel_interval_static1($sA, $sb)
+        t4 = @belapsed gauss_seidel_interval_static2($sA, $sb)
+        t5 = @belapsed gauss_seidel_contractor($A, $b)
```
dpsanders: Can you try this one with the static vectors too?

perf/linear_eq.jl
```diff
@@ -0,0 +1,40 @@
...
+        df[Symbol("n = $n")] = [t1, t2, t3, t4, t5]
```
dpsanders: Why do you need `Symbol` here? Can't you just use a string?

dpsanders: Why don't you in fact just make `n` another column?

perf/linear_eq_tabular.txt
```diff
@@ -0,0 +1,22 @@
...
+│ 11  │ n = 11   │ 0.000465017 │ 0.000489114 │ 0.00585246  │ 0.000473425 │ 0.000447176 │
+│ 12  │ n = 12   │ 0.00055403  │ 0.000587152 │ 0.0110894   │ 0.000573395 │ 0.00054274  │
+│ 13  │ n = 13   │ 0.000659994 │ 0.000692402 │ 0.0160662   │ 0.000686098 │ 0.000643391 │
+│ 14  │ n = 14   │ 0.000777718 │ 0.000793626 │ 0.0184595   │ 0.000800443 │ 0.000767576 │
+│ 15  │ n = 15   │ 0.000910174 │ 0.000952207 │ 0.0239422   │ 0.000972855 │ 0.000899402 │
+│ 16  │ n = 16   │ 0.00107785  │ 0.00111443  │ 0.0303295   │ 0.00113467  │ 0.0010867   │
+│ 17  │ n = 17   │ 0.00122326  │ 0.00125759  │ 0.0398127   │ 0.00134961  │ 0.00125675  │
+│ 18  │ n = 18   │ 0.00167176  │ 0.00159463  │ 0.043795    │ 0.00181409  │ 0.00162584  │
+│ 19  │ n = 19   │ 0.0018632   │ 0.00185617  │ 0.0549107   │ 0.00208529  │ 0.00196587  │
+│ 20  │ n = 20   │ 0.0020604   │ 0.00210595  │ 0.0635598   │ 0.00232704  │ 0.00212374  │
```
dpsanders: Actually `n` seems to be redundant since it's just the same as the row.

src/linear_eq.jl
```diff
@@ -0,0 +1,203 @@
...
+Keyword `precondition` to turn preconditioning off.
+Eldon Hansen and G. William Walster : Global Optimization Using Interval Analysis - Chapter 5 - Page 115
+"""
+function gauss_seidel_interval!{T}(x::Array{Interval{T}}, A::Matrix{Interval{T}}, b::Array{Interval{T}}; precondition=true, maxiter=100)
+
+    precondition && ((A, b) = preconditioner(A, b))
+
+    n = size(A, 1)
+
+    for iter in 1:maxiter
```
dpsanders: You should add `@inbounds` to the outer `for` loop in each of these functions.

dpsanders:
You should squash down the implementations as I suggested; I think you should leave all of them for now.

Could you please add the example from the Jaulin book in the `examples` folder? 

Actually you should keep a separate repo, e.g. in your GitHub account or in JuliaIntervals, with a Jupyter notebook with usage examples and explanation.

src/linear_eq.jl
```diff
@@ -0,0 +1,203 @@
...
+end
+
+"""
+Iteratively solves the system of interval linear
+equations and returns the solution set. Uses the
+Gauss-Seidel method (Hansen-Sengupta version) to solve the system.
+Keyword `precondition` to turn preconditioning off.
+Eldon Hansen and G. William Walster : Global Optimization Using Interval Analysis - Chapter 5 - Page 115
+"""
+function gauss_seidel_interval_static1!{T, N}(x::MVector{N, Interval{T}}, A::SMatrix{N, N, Interval{T}}, b::SVector{N, Interval{T}}; precondition=true, maxiter=100)
```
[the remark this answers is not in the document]
dpsanders: (Or no type annotations at all)

perf/linear_eq.jl
eeshan9815: `gauss_seidel_contractor` with static vectors? Okay

perf/linear_eq.jl
eeshan9815:
```
julia> df["test"] = [1, 2, 3, 4]
ERROR: MethodError: no method matching setindex!(::DataFrames.DataFrame, ::Array{Int64,1}, ::String)
```
The `DataFrames` package takes a symbol as the index.

perf/linear_eq_tabular.txt
eeshan9815: The row numbers need not be sequential due to the stack operations, That's why I kept the column with the value of n

eeshan9815:
@dpsanders Is this accurate?

src/linear_eq.jl
```diff
@@ -0,0 +1,217 @@
...
+
+end
+
+function gauss_seidel_interval(A::AbstractMatrix, b::AbstractArray; precondition=true, maxiter=100)
+
+    n = size(A, 1)
+    x = fill(-1e16..1e16, n)
+
+    if (typeof(b) <: SArray || typeof(b) <: MArray)
+        x = MVector{n}(x)
```
eeshan9815: @dpsanders According to the julia docs (https://docs.julialang.org/en/stable/manual/performance-tips/#Avoid-changing-the-type-of-a-variable-1), explicit conversions can be used to avoid optimization loss. Can `x` be changed to `MVector{n}` in this way?

eeshan9815: I have added the new benchmarks as well, they seem similar to the earlier ones.

dpsanders: No, you should not do this in the case of an `SArray` because it changes the type of `x`, thereby introducing a type instability. (I think you misread the Julia docs, although maybe they aren't very clear.)

src/linear_eq.jl
```diff
@@ -0,0 +1,217 @@
...
+            for j in 1:n
+                (i == j) || (Y -= A[i, j] * x[j])
+            end
+            Z = extended_div(Y, A[i, i])
+            x[i] = hull((x[i] ∩ Z[1]), x[i] ∩ Z[2])
+        end
+    end
+    x
+end
+#
```
dpsanders: All of this should be deleted instead of commented out.
It can be recovered from the git history if necessary.

src/linear_eq.jl
```diff
@@ -0,0 +1,217 @@
...
+#     x
+# end
+
+function gauss_seidel_contractor(A::AbstractMatrix, b::AbstractArray; precondition=true, maxiter=100)
+
+    n = size(A, 1)
+    x = fill(-1e16..1e16, n)
+
+    if (typeof(b) <: SArray || typeof(b) <: MArray)
+        x = MVector{n}(x)
```
dpsanders: I don't understand why this is necessary. (I don't think it is.)

eeshan9815: So `x` can remain an ordinary `Vector` even if `A` and `b` are `SVector`s and `MVector`s?

src/linear_eq.jl
eeshan9815: Okay, got it.
What do the docs mean when they say
> There are several possible fixes:
>  Use an explicit conversion: `x = oneunit(T)`

src/linear_eq.jl
dpsanders: No, `x` should be an `SVector` if `A` and `b` are `SVector`s I think?
Maybe I'm misunderstanding.

src/linear_eq.jl
```diff
@@ -215,3 +75,77 @@ function gauss_seidel_contractor!(x::AbstractArray, A::AbstractMatrix, b::Abstra
     end
     x
 end
+
+function gauss_elimination_interval(A::AbstractMatrix, b::AbstractArray; precondition=true)
+
+    n = size(A, 1)
+    x = fill(-1e16..1e16, n)
+
+    x = gauss_seidel_interval!(x, A, b, precondition=precondition)
```
dpsanders: Should be `gauss_elimination_interval!`  (not `seidel`)

src/linear_eq.jl
```diff
@@ -215,3 +75,77 @@ function gauss_seidel_contractor!(x::AbstractArray, A::AbstractMatrix, b::Abstra
...
+
+    p .∩ x
+end
+
+function gauss_elimination_interval1(A::AbstractMatrix, b::AbstractArray; precondition=true)
+
+    n = size(A, 1)
+    x = fill(-1e16..1e16, n)
+
+    x = gauss_seidel_interval!(x, A, b, precondition=precondition)
```
dpsanders: `elimination`

src/linear_eq.jl
```diff
@@ -215,3 +75,77 @@ function gauss_seidel_contractor!(x::AbstractArray, A::AbstractMatrix, b::Abstra
...
+
+    return x
+end
+"""
+Iteratively solves the system of interval linear
+equations and returns the solution set. Uses the
+Gauss-Seidel method (Hansen-Sengupta version) to solve the system.
```
dpsanders: This text is wrong.

src/linear_eq.jl
```diff
@@ -215,3 +75,77 @@ function gauss_seidel_contractor!(x::AbstractArray, A::AbstractMatrix, b::Abstra
...
```
dpsanders: This text is wrong.

src/linear_eq.jl
dpsanders: Probably you should use `similar` to make a vector of the correct type.

src/linear_eq.jl
eeshan9815: I'm sorry I don't understand?

dpsanders:
This and your other PRs need rebasing on current master, sorry!

dpsanders:
There is still (or again) a conflict in `test/runtests.jl`.

dpsanders:
(sorry, wrong button)

eeshan9815:
Oh that occurred due to the merging of #70. I have rebased again.

dpsanders:
@eeshan9815 Tests are still failing. Do you know what is happening?

eeshan9815:
Tests fail as it uses `extended_div` which is not in the latest tagged version of `IntervalArithmetic`. 
Tests pass locally if the current master is used.

dpsanders:
Ah great, thanks

eeshan9815:
I have redefined `\` for `Interval`s. Currently, it uses the `gauss_elimination_interval` method with `precondition=true`.
This is due to the advantages of preconditioning such as reducing dependence mentioned in various books such as Global Optimisation using Interval Analysis by Eldon Hansen.

eeshan9815:
Since `\` is same as `gauss_elimination_interval`, for which tests exist already, is there a need to define tests for `\` separately?
Also, comparisions exist in the `perf/linear_eq_tabular.txt` file.

dpsanders:
Right, but the tests are also for future-proofing, eg if we change the implementation of \ again. So you can just add a couple of tests to make sure the output of \ and gaussian elim are the same.

dpsanders:
@eeshan9815 Did you actually check that the definition of `\` in the package is being called? It doesn't seem to be. You need to `import Base: \`.

dpsanders:
On the example in the testset "Linear Equations", the new definition seems to be worse than using Base's `\`.

eeshan9815:
I have imported `Base.\` in `IntervalRootFinding.jl`and it seems to be using `gauss_elimination_interval` instead of `Base.\`
Also, preconditioning widens the solution set by a little bit. It's a trade-off mentioned in the Hansen book as well.

dpsanders:
Hmm, I see, sorry. I don't know why it didn't work for me, then.

What's the advantage of preconditioning, then? I thought it was supposed to narrow, not widen, the result.

dpsanders:
After checking out your PR, I get
```
julia> A = [4..5 -1..1 1.5..2.5; -0.5..0.5 -7.. -5 1..2; -1.5.. -0.5 -0.7.. -0.5 2..3]
3×3 Array{IntervalArithmetic.Interval{Float64},2}:
 [4, 5]            [-1, 1]                    [1.5, 2.5]
    [-0.5, 0.5]   [-7, -5]                [1, 2]
    [-1.5, -0.5]       [-0.700001, -0.5]  [2, 3]

julia> b = [3..4, 0..2, 3..4]
3-element Array{IntervalArithmetic.Interval{Float64},1}:
 [3, 4]
 [0, 2]
 [3, 4]

julia> A \ b
3-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-1.81928, 1.16873]
 [-0.41407, 1.72523]
  [0.700232, 3.42076]

julia> @which A \ b
\(A::AbstractArray{T,2} where T, B::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T) in Base.LinAlg at linalg/generic.jl:820
```
which shows that the `\` from `Base` is being used. Not sure why.

dpsanders:
Ignore me, I seem not to have managed to checkout the latest version, apologies.

dpsanders:
No, don't ignore me -- you defined `\` in `test/linear_eq.jl` instead of in the source code!

test/linear_eq.jl
```diff
@@ -0,0 +1,20 @@
...
+    for i in 1:length(A)
+        for precondition in (false, true)
+            for f in (gauss_seidel_interval, gauss_seidel_contractor, gauss_elimination_interval, \)
+                @test all(x[i] .⊆ f(A[i], b[i]))
+            end
+        end
+    end
+end
+
+\(A::StaticMatrix{Interval{T}}, b::StaticArray{Interval{T}}; kwargs...) where T = gauss_elimination_interval(A, b, kwargs...)
```
dpsanders: These should be moved to the source code.

I think you can define a single method if you use `AbstractMatrix` and `AbstractVector`.

eeshan9815: >These should be moved to the source code.

I'm terribly sorry, I didn't realize.

>I think you can define a single method if you use AbstractMatrix and AbstractVector

I tried `AbstractArray`s earlier, but it doesn't work since it matches its type with the `\` for `StaticArray`s or `Array`s instead.

>What's the advantage of preconditioning, then? I thought it was supposed to narrow, not widen, the result

Preconditioning doesn't always give a better result. Specifically, for [M-matrices](https://en.wikipedia.org/wiki/M-matrix), the non-preconditioned version computes the exact hull whereas preconditioning causes some widening.

dpsanders:
No problem, thanks!

dpsanders:
Running `gauss_seidel` seems to modify something it shouldn't? The result of `A \ b` shouldn't change in the following session:

```jl
julia> IntervalRootFinding.gauss_elimination_interval(A, b)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-130.228, 167.728]
  [-60.0001, 267.273]

julia> IntervalRootFinding.gauss_elimination_interval(A, b, precondition=false)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-120, 90]
 [-60, 240]

julia> IntervalRootFinding.gauss_seidel_interval(A, b, precondition=false)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-5e+15, 5.00001e+15]
 [-1e+16, 1e+16]

julia> A
2×2 Array{IntervalArithmetic.Interval{Float64},2}:
 [2, 3]  [0, 1]
 [1, 2]  [1, 3]

julia> b
2-element Array{IntervalArithmetic.Interval{Float64},1}:
   [0, 120]
 [-60, 240]

julia> A \ b
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-1e+16, 1e+16]
 [-1e+16, 1e+16]
```

dpsanders:
I think this needs more tests. E.g. the `gauss_seidel` methods give huge pointless answers for the single test case.

src/linear_eq.jl
```diff
@@ -0,0 +1,156 @@
...
+        x¹ = copy(x)
+        for i in 1:n
+            Y = b[i]
+            for j in 1:n
+                (i == j) || (Y -= A[i, j] * x[j])
+            end
+            Z = extended_div(Y, A[i, i])
+            x[i] = hull((x[i] ∩ Z[1]), x[i] ∩ Z[2])
+        end
+        if all(x .== x¹)
```
dpsanders: This should probably use a tolerance instead.
Can you write this as `all(==, x, x1)`?

dpsanders:
@eeshan9815 There seems to be a real test failure here.

eeshan9815:
The test failure appears to be in a test for 2D roots, giving error messages similar to the ones faced when the `SVector` was initially made a data member of `IntervalBox`

eeshan9815:
The incorrect modification issue has also been resolved.
```
julia> IntervalRootFinding.gauss_elimination_interval(A, b)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-130.228, 167.728] 
  [-60.0001, 267.273]

julia> IntervalRootFinding.gauss_elimination_interval(A, b, precondition=false)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-120, 90]
 [-60, 240]

julia> IntervalRootFinding.gauss_seidel_interval(A, b, precondition=false)
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-120.001, 90.0001] 
  [-60.0001, 240.001]

julia> A
2×2 Array{IntervalArithmetic.Interval{Float64},2}:
 [2, 3]  [0, 1]
 [1, 2]  [2, 3]

julia> b
2-element Array{IntervalArithmetic.Interval{Float64},1}:
  [0, 120]
 [60, 240]

julia> A \ b
2-element Array{IntervalArithmetic.Interval{Float64},1}:
 [-130.228, 167.728] 
  [-60.0001, 267.273]
```

dpsanders:
@eeshan9815 Tests are now passing. Is this ready?

eeshan9815:
Yes, I think so. I added 88 tests (mostly randomly generated) to make sure the methods are working.

dpsanders:
Thanks very much!

PR #68, opened 2018-05-18
Add solvers for quadratic interval equations

eeshan9815:


eeshan9815:
ping @dpsanders

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+    push!(L, x0)
+end
+
+
+"""
+Function to solve a quadratic equation where the coefficients are intervals.
+Returns an array of intervals of the roots.
+Eldon Hansen and G. William Walster : Global Optimization Using Interval Analysis - Chapter 8
+"""
+function quadratic_interval{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
```
dpsanders: It's maybe not clear from the function name that this solves a quadratic.
Could be e.g. `quadratic_roots`

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
+"""
+Helper function for quadratic_interval
+"""
+function quadratic_helper(a::Number, b::Number, c::Number, L::AbstractArray)
```
dpsanders: If this modifies `L`, then it should be 

```
quadratic_helper!(L, a, b, c)
```

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    Δ = b*b - 4*a*c
+
+    Δ < 0 && return
+
+    Δ = sqrt(Δ)
+
+    if (b >= 0)
+        x0 = -0.5 * (b + Δ)
```
dpsanders: Do the calculations in this function not need directed rounding?

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    else
+        x0 = -0.5 * (b - Δ)
```
dpsanders: Why is this distinction necessary?

src/quadratic.jl
dpsanders: These two functions implement a non-obvious algorithm, so I would suggest adding at least some hints as to what's going on, even though it's explained in the book.

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    if (x0 == 0)
+        x1 = x0
```
dpsanders: This seems to be redundant.

dpsanders: Probably just `push!(L, x0)` here.

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    else
+        x1 = c / x0
+        x0 = x0 / a
+        push!(L, x1)
```
dpsanders: And `push!(L, x0, x1)` here.
That way it's clearer that there is a single root in the previous case and 2 here.

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    L = T[]
+    R = Interval{T}[]
+
+    quadratic_helper(a.lo, b.lo, c.lo, L)
+    quadratic_helper(a.hi, b.hi, c.hi, L)
+    quadratic_helper(a.lo, b.hi, c.lo, L)
+    quadratic_helper(a.hi, b.lo, c.hi, L)
+
+    if (length(L) == 8)
```
dpsanders: Pretty unclear what's going on here.

src/quadratic.jl
```diff
@@ -0,0 +1,66 @@
...
+        push!(L, -∞)
+    end
+
+    if (a.lo < 0 || (a.lo == 0 && b.lo == 0) || (a.lo == 0 && b.lo == 0 && c.lo ≤ 0))
+        push!(L, ∞)
+    end
+
+    sort!(L)
+
+    for i in 1:2:length(L)
```
dpsanders: This is also strange.

test/quadratic.jl
```diff
@@ -0,0 +1,14 @@
+using IntervalArithmetic, IntervalRootFinding
+
+@testset "Quadratic Interval Equations" begin
+    a = [interval(1, 5), interval(-1, 1), interval(1, 2), ]
+    b = [interval(2, 4), interval(-2, 2), interval(1, 3), ]
+    c = [interval(0, 2), interval(-1, 1), interval(2, 6), ]
+    x = [[interval(-4, -2), interval(-0, -0)], [interval(-∞, -1), interval(-1, -1), interval(-1, ∞)], [interval(-2, -1)], ]
+    l = [2, 3, 1, ]
+    for i in 1:length(a)
+            roots = quadratic_interval(a[i], b[i], c[i])
```
dpsanders: I think it would be clearer to have each test case as a separate object, e.g. a list of objects of the form

`( (a, b, c), [roots] )`

or even make an actual `struct` for each test case.

Note that you don't need a separate `l` if you have the root vectors.

src/quadratic.jl
eeshan9815: According to the Hansen book (Page 181), this procedure leads to minimisation of rounding errors.

src/quadratic.jl
eeshan9815: They can be `Float`s, so I think interval arithmetic can be used to bound rounding errors. Should I update to include that?

src/quadratic.jl
eeshan9815: The length of `L` can only be 8 if all `a`, `b` and `c` are degenerate intervals, otherwise it can only reach 6 since an interval quadratic equation can only return a maximum of 3 intervals.
If degenerate intervals are provided, 2 such intervals are redundant and can be removed by keeping only the first four entries of `L`.

src/quadratic.jl
eeshan9815: Sorted the `Float`s and generated `Interval`s from them as mentioned in the Hansen book.

eeshan9815:
TODO:Documentation

eeshan9815:
ping @dpsanders

src/quadratic.jl
```diff
@@ -0,0 +1,72 @@
+"""
+Helper function for quadratic_interval that computes roots of a
+real quadratic using interval arithmetic to bound rounding errors.
+"""
+function quadratic_helper!{T}(a::Interval{T}, b::Interval{T}, c::Interval{T}, L::Array{Interval{T}})
+
+    Δ = b*b - 4*a*c
```
Kolaru: Is there a reason to use `b*b` rather than `b^2` ? If there is it would be good to add a small comment explaining why, since the two expressions are not equivalent in general.

eeshan9815: `b^2` should be better here as it gives better benchmarks

PR #69, opened 2018-05-22
Automatic evaluation of slope expansions

eeshan9815:
ping @dpsanders 
Reference - Dietmar Ratz - An Optimized Interval Slope Arithmetic and its Application (http://www2.math.uni-wuppertal.de/~xsc/preprints/rep9604.pdf)


lbenet:
Naive question: this seems very close to automatic differentiation. Can you explain the difference?

eeshan9815:
Slope Expansion usually returns much smaller intervals as compared to AD, especially as the interval size increases. 
Given below are the benchmarking results run on seven functions on the interval `0.75..1.75`, and the slope is expanded about the `mid` of the interval.
```
│ Row │ Function │ Automatic Differentiation │ Slope Expansion       │
├─────┼──────────┼───────────────────────────┼───────────────────────┤
│ 1   │ f1       │ [-5.44573, 0.886249]      │ [-2.79917, 0.0521429] │
│ 2   │ f2       │ [-87.6875, 77.0625]       │ [-43.875, 38.25]      │
│ 3   │ f3       │ [-0.474861, 0.787211]     │ [-0.159199, 0.432842] │
│ 4   │ f4       │ [-2.97001, 21.0701]       │ [0.04, 0.326667]      │
│ 5   │ f5       │ [2.63258, 74.8333]        │ [6.03135, 33.2205]    │
│ 6   │ f6       │ [-94.5871, 115.135]       │ [-38.9908, 65.5595]   │
│ 7   │ f7       │ [-279.639, 167.667]       │ [-146.852, 67.0665]   │
```

dpsanders:
As a general comment, it would be useful to provide more comments / documentation about the basic idea of the algorithm etc. And a link to the PDF if it's freely available.

dpsanders:
Why do slopes usually perform better?

eeshan9815:
There's a theorem given in the book by Kearfott (Theorem 1.10) which proves that the slope expansion always gives a tighter bound than the derivative even when w(x) approaches 0.

dpsanders:
Sorry, forgot to send the review comments :/

perf/slopes.jl
```diff
@@ -0,0 +1,57 @@
...
+    push!(f, x->((log(x + 1.25) - 0.84x) ^ 2))
+    push!(f, x->(0.02x^2 - 0.03exp(-(20(x - 0.875))^2)))
+    push!(f, x->(exp(x^2)))
+    push!(f, x->(x^4 - 12x^3 + 47x^2 - 60x - 20exp(-x)))
+    push!(f, x->(x^6 - 15x^4 + 27x^2 + 250))
+    df = DataFrame()
+    df[:Method] = ["Automatic Differentiation", "Slope Expansion"]
+    for n in 1:length(f)
+
+        t1 = ForwardDiff.derivative(f[n], 0.75..1.75)
```
dpsanders: Make the interval a variable since you use it twice (here and in following line).

perf/slopes.jl
```diff
@@ -0,0 +1,57 @@
...
+        t2 = slope(f[n], 0.75..1.75, 1.25)
```
dpsanders: Use `mid` for the last argument. 
You should (almost) *never* have arbitrary numbers floating around.

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
+# Reference : Dietmar Ratz - An Optimized Interval Slope Arithmetic and its Application
+using IntervalArithmetic, ForwardDiff
+import Base: +, -, *, /, ^, sqrt, exp, log, sin, cos, tan, asin, acos, atan
+
+function slope(f::Function, x::Interval, c::Real)
+    try
```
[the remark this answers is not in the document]
eeshan9815: In case there's some elementary method that the algorithm can't build from other methods, example `sinh` (for now), then the function shouldn't return a `MethodError` but return the bound generated by automatic differentiation.

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+        f(SlopeVar(x, c)).fs
+    catch y
+        if isa(y, MethodError)
+            ForwardDiff.derivative(f, x)
+        end
+    end
+end
+
+struct SlopeType
```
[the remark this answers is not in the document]
eeshan9815: The midpoint is a floating point, but I have kept that as an `Interval` to bound rounding errors.

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+    fx::Interval
```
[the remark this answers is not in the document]
eeshan9815: Okay, I'll update accordingly.

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+    fc::Interval
```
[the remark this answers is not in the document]
eeshan9815: Okay, basically make the `struct` as `Slope{T}`?

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+    fs::Interval
+end
+
+function SlopeConst(c::Union{Real, Interval})
```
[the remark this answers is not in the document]
eeshan9815: Okay, I'll update.

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+            ForwardDiff.derivative(f, x)
+        end
+    end
+end
+
+struct Slope{T}
+    x::Interval{T}
+    c::Interval{T}
+    s::Interval{T}
+    Slope{T}(a, b, c) where T = new(a, b, c)
```
dpsanders: This is the default constructor that will be created automatically if there are no inner constructors.

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+    Slope{T}(c) where T = Slope{T}(c, c, 0)
```
dpsanders: This can be an outer constructor.

Inner constructors are necessary only if you need to impose something particular when an object is created (eg that something cannot be negative).

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+    catch y
+        if isa(y, MethodError)
+            ForwardDiff.derivative(f, x)
+        end
+    end
+end
+
+struct Slope{T}
+    x::Interval{T}
+    c::Interval{T}
```
dpsanders: This should not be an Interval I believe.

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+    s::Interval{T}
+    Slope{T}(a, b, c) where T = new(a, b, c)
+    Slope{T}(c) where T = Slope{T}(c, c, 0)
+end
+
+function slope_var(v::Real)
+    Slope{Float64}(v, v, 1)
```
dpsanders: Why are you imposing Float64?

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+    u.c
+end
+
+function slope(u::Slope)
+    u.s
+end
+
+function +(u::Slope, v::Slope)
+    Slope{Float64}(u.x + v.x, u.c + v.c, u.s + v.s)
+end
```
dpsanders: Do not restrict the type to Float64 anywhere. Everything should be generic. You shouldn't need to specify that.

dpsanders: In fact, rather you should add some tests using eg BigFloat.

src/slopes.jl
eeshan9815: Won't the `Real` need to bound rounding errors?

dpsanders: That depends how it's used. I'm not sure.

eeshan9815: Interval arithmetic would be used everywhere if `c` is an `Interval` and that should help bound rounding errors, right?

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+
+struct Slope{T}
+    x::Interval{T}
+    c::Interval{T}
+    s::Interval{T}
+end
+
+Slope{T}(c) where T = Slope{T}(c, c, 0)
+
+function slope_var{T}(v::T)
```
dpsanders: Please use the new syntax with where instead of this syntax everywhere for parameterised functions.

src/slopes.jl
```diff
@@ -0,0 +1,198 @@
...
+
+function *{T}(u::Slope{T}, v::Slope{T})
+    Slope{T}(u.x * v.x, u.c * v.c, u.s * v.c + u.x * v.s)
+end
+
+function /{T}(u::Slope{T}, v::Slope{T})
+    Slope{T}(u.x / v.x, u.c / v.c, (u.s - (u.c / v.c) * v.s) / v.x)
+end
+
+function +{T}(u::Union{Interval{T}, Number}, v::Slope{T})
```
dpsanders: Do these functions really need to be parameterised? And do they need type annotations for the first parameter?

eeshan9815: The functions do need to be parameterized since I'm making constructor calls in the functions that need the `convert` method to be called on them, and that would be done only if the parameterized constructor is called.
(Reference - https://docs.julialang.org/en/stable/manual/types/#Parametric-Composite-Types-1)
```
julia> Slope(1..2, 1..2, 1.5)
ERROR: MethodError: no method matching Slope(::IntervalArithmetic.Interval{Float64}, ::IntervalArithmetic.Interval{Float64}, ::Float64)
Closest candidates are:
  Slope(::IntervalArithmetic.Interval{T}, ::IntervalArithmetic.Interval{T}, ::IntervalArithmetic.Interval{T}) where T at none:2
Stacktrace:
 [1] macro expansion at /home/eeshan/.julia/v0.6/Atom/src/repl.jl:118 [inlined]
 [2] anonymous at ./<missing>:?

julia> Slope{Float64}(1..2, 1..2, 1.5)
Slope{Float64}([1, 2], [1, 2], [1.5, 1.5])
```

dpsanders: You can just add a constructor Slope(a, b, c) that does the conversion.

eeshan9815: And for the type annotation, the behaviour of a function, say `+`, is different when adding two `Slope` types as compared to adding a `Slope` to an `Interval` or a  `Float`.
That's why I thought type annotations was necessary. Or should I attach annotations only for the function that adds two `Slope`s?

src/slopes.jl
eeshan9815: Oh okay, I was using the syntax that was prevalent in the repo already. I will change accordingly.

eeshan9815:
@dpsanders Updated the PR with the multidimensional slope function

perf/slopes.jl
```diff
@@ -0,0 +1,116 @@
...
+    end
+    df1 = stack(df, a)
+    dfnew = unstack(df1, :variable, :Method, :value)
+    dfnew = rename(dfnew, :variable => :Function)
+    println(dfnew)
+    dfnew
+end
+
+struct SlopesMulti
+    f::Function
```
dpsanders: These are all abstract types again...

perf/slopes.jl
```diff
@@ -0,0 +1,116 @@
...
+    x::IntervalBox
+    c::Vector
+    sol::Matrix{Interval}
+end
+
+function benchmark_multi()
```
dpsanders: Could you please move all the multidimensional stuff to a new PR?

test/slopes.jl
```diff
@@ -0,0 +1,93 @@
...
+    push!(rts, Slopes(x->(x^6 - 15x^4 + 27x^2 + 250), s, mid(s), interval(-146.9, 67.1)))
+    push!(rts, Slopes(x->(atan(cos(tan(x)))), s, mid(s), interval(1, 2)))
+    push!(rts, Slopes(x->(asin(cos(acos(sin(x))))), s, mid(s), interval(1.36, ∞)))
+
+    for i in 1:length(rts)
+            @test slope(rts[i].f, rts[i].x, rts[i].c) ⊆ rts[i].sol
+    end
+end
+
+@testset "Automatic slope expansion(BigFloat)" begin
```
dpsanders: You should be able to unify these tests with the `Float64` ones, so that there's only one function that tests both types.

dpsanders: You should always try to avoid duplicating code: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

perf/slopes.jl
dpsanders: Maybe it doesn't matter so much for benchmarking.

src/slopes.jl
```diff
@@ -0,0 +1,228 @@
...
+    try
+        f(slope_var(x, c)).s
+    catch y
+        if isa(y, MethodError)
+            ForwardDiff.derivative(f, x)
+        end
+    end
+end
+
+function slope(f::Function, x::Union{IntervalBox, SVector}, c::AbstractVector = mid.(x)) # multidim
```
dpsanders: So you don't have a separate type representing a multidimensional slope? Probably you should, and it should probably be an `SVector` or `IntervalBox`?

perf/slopes.jl
eeshan9815: Yeah, that's why I didn't care much because this script is only to be run for benchmarking, and doesn't influence the results.

perf/slopes.jl
eeshan9815: Okay, I will

src/slopes.jl
eeshan9815: I will have to write overloading rules for all of them again, and most implementations use the result of the 1D `slope` function to construct the matrix for the multidimensional version. So it will be really complicated to have a new type for multidimensional slope. The function returns an `SArray` as is.

eeshan9815:
I have removed the multidim slope function from this PR and unified tests for `Float64` and `BigFloat`

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+        f(slope_var(x, c)).s
+    catch y
+        if isa(y, MethodError)
+            ForwardDiff.derivative(f, x)
+        end
+    end
+end
+
+struct Slope{T}
+    x::Interval{T}
```
dpsanders: Please comment what these variables are. Or you could rename them to reflect their meaning.

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
...
+    c::Interval{T}
+    s::Interval{T}
+end
+
+Slope(c) = Slope(c, c, 0)
+Slope(a, b, c) = Slope(promote(convert(Interval, a), b, c)...)
+
+function slope_var(v::Number)
```
dpsanders: What does this do?

src/slopes.jl
```diff
@@ -0,0 +1,199 @@
+# Reference : Dietmar Ratz - An Optimized Interval Slope Arithmetic and its Application
+using IntervalArithmetic, ForwardDiff
+import Base: +, -, *, /, ^, sqrt, exp, log, sin, cos, tan, asin, acos, atan
+import IntervalArithmetic: mid, interval
+
+function slope(f::Function, x::Interval, c::Number)
+    try
```
dpsanders: I think you should remove the `try` blocks and let the function fail if the function is non-slopable.

test/slopes.jl
```diff
@@ -0,0 +1,30 @@
...
+        push!(rts, Slopes(x->((log(x + 1.25) - 0.84x) ^ 2), s, mid(s), interval(T(-0.16), T(0.44))))
+        push!(rts, Slopes(x->(0.02x^2 - 0.03exp(-(20(x - 0.875))^2)), s, mid(s), interval(T(0.03), T(0.33))))
+        push!(rts, Slopes(x->(exp(x^2)), s, mid(s), interval(T(6.03), T(33.23))))
+        push!(rts, Slopes(x->(x^4 - 12x^3 + 47x^2 - 60x - 20exp(-x)), s, mid(s), interval(T(-39), T(65.56))))
+        push!(rts, Slopes(x->(x^6 - 15x^4 + 27x^2 + 250), s, mid(s), interval(T(-146.9), T(67.1))))
+        push!(rts, Slopes(x->(atan(cos(tan(x)))), s, mid(s), interval(T(1), T(2))))
+        push!(rts, Slopes(x->(asin(cos(acos(sin(x))))), s, mid(s), interval(T(1.36), T(∞))))
+
+        for i in 1:length(rts)
+                @test slope(rts[i].f, rts[i].x, rts[i].c) ⊆ rts[i].sol
```
dpsanders: Wrong indent

test/slopes.jl
```diff
@@ -0,0 +1,30 @@
...
+    f::Function
+    x::Interval{T}
+    c::T
+    sol::Interval{T}
+end
+
+@testset "Automatic slope expansion" begin
+    for T in [Float64, BigFloat]
+        s = interval(T(0.75), T(1.75))
+        rts = Slopes{T}[]
```
dpsanders: I think it's confusing to call this `rts`.

dpsanders:
Please add some documentation.

dpsanders:
@eeshan9815 Is this ready?

eeshan9815:
Yes, definitely.

dpsanders:
Sorry, needs a rebase now.

eeshan9815:
Done!

dpsanders:
Thanks!

Issue #44, opened 2018-03-12
Add hard test cases


dpsanders:
There are some nice ones in

https://www.sciencedirect.com/science/article/pii/S0377042700007111

with results we can compare with.

gwater:
For your information: I'm preparing a pull request with implementations of some examples from Smiley and Chun's article. So far, all examples pass (with the Newton contractor) although some (specifically Example 5.5) take upwards of 30 minutes on my machine.

dpsanders:
Great, thanks!

dpsanders:
Done in #70.

PR #70, opened 2018-05-25
Add some hard test cases from Smiley and Chun (2001)

gwater:
Addresses plans in #44 to implement some more difficult test cases by implementing 4 examples from [M. W. Smiley and C. Chun, J. Comput. Appl. Math. **137**, 293 (2001)](https://doi.org/10.1016/S0377-0427(00)00711-1).

Currently all examples pass (using the Newton method). However, Example 5.5 takes ca 40 minutes on my machine. Therefore it might not be practicable to include the test cases in the default set of tests.

Example 5.5 might be useful for benchmarks and checking for performance regressions.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -6,4 +6,4 @@ using Base.Test
 include("roots.jl")
 include("newton1d.jl")
 include("quadratic.jl")
-#include("test_smiley.jl")
+include("test_smiley.jl")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -6,3 +6,4 @@ using Base.Test
 include("roots.jl")
 include("newton1d.jl")
 include("quadratic.jl")
+#include("test_smiley.jl")
--- a/test/smiley_examples.jl
+++ b/test/smiley_examples.jl
@@ -126,7 +126,7 @@ const known_roots = [
     IntervalBox(-1.34030 ± 1e-5,  1.00134 ± 1e-5),
     IntervalBox( 1.34030 ± 1e-5,  0.99866 ± 1e-5),
     IntervalBox( 1.34298 ± 1e-5, -0.99866 ± 1e-5),
-    # following three roots are not reported accurately
+    # following three roots are not reported precisely
     IntervalBox(-7e-10 ± 1e-10,  1 ± 1e-5),
     IntervalBox(-7e-10 ± 1e-10, -1 ± 1e-5),
     IntervalBox(-5e2 ± 1, -5e2 ± 1)
--- a/test/smiley_examples.jl
+++ b/test/smiley_examples.jl
@@ -1,0 +1,220 @@
+# examples from
+# M. W. Smiley and C. Chun, J. Comput. Appl. Math. **137**, 293 (2001).
+# https://doi.org/10.1016/S0377-0427(00)00711-1
+
+module SmileyExample22
+
+using IntervalArithmetic
+using StaticArrays
+
+const title = "Smiley and Chun (2001), Example 2.2"
+
+f(x) = SVector(
+    x[1]^2 + 4 * x[2]^2 - 4,
+    x[2] * (x[1] - 1.995) * (x[2] - x[1]^2) * (x[2] - x[1] + 1)
+)
+
+# contains all 8 reported roots
+const region = IntervalBox(-3..3, -3..3)
+
+# only three roots are reported explicitely:
+# (2, 0) and (1.995, ±0.071)
+
+end
+
+module SmileyExample52
+
+using IntervalArithmetic
+using StaticArrays
+
+const title = "Smiley and Chun (2001), Example 5.2"
+
+const c0  =  1
+const a0  =  0.5
+const b0  =  0.5
+const d01 =  0.2
+const d02 = -0.7
+const r1  =  1
+const r2  =  2
+const m   =  3
+const θs  = [i * pi / m for i in 1:m]
+_v(θ) = -(a0 * cos(θ) + b0 * sin(θ)) / c0
+const vs = _v.(θs)
+
+_a(θ, v) = b0 * v - c0 * sin(θ)
+const as = _a.(θs, vs)
+_b(θ, v) = c0 * cos(θ) - a0 * v
+const bs = _b.(θs, vs)
+_c(θ) = a0 * sin(θ) - b0 * cos(θ)
+const cs = _c.(θs)
+_d(c) = d01 * c / c0
+const ds = _d.(cs)
+
+f(x) = SVector(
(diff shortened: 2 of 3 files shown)


test/smiley_examples.jl
```diff
@@ -0,0 +1,220 @@
...
+
+    IntervalBox(-0.499927 ± 1e-6,  0.865900 ± 1e-6,  0.017013 ± 1e-6),
+    IntervalBox(-0.360640 ± 1e-6, -0.624646 ± 1e-6,  0.692643 ± 1e-6),
+    IntervalBox( 0.082249 ± 1e-6, -0.962075 ± 1e-6, -0.260086 ± 1e-6),
+    IntervalBox( 0.085220 ± 1e-6,  0.367221 ± 1e-6, -0.926221 ± 1e-6),
+    IntervalBox( 0.453788 ± 1e-6,  0.785984 ± 1e-6, -0.419886 ± 1e-6),
+
+    IntervalBox( 0.464511 ± 1e-6, -0.804557 ± 1e-6,  0.370022 ± 1e-6),
+    IntervalBox( 0.511026 ± 1e-6, -0.300000 ± 1e-6, -0.805513 ± 1e-6),
+    # the following two roots are suspect, first column probably reported in error
```
dpsanders: Do you mean that the error is in the original paper?

test/smiley_examples.jl
```diff
@@ -0,0 +1,220 @@
...
+
+# contains all 7 reported roots
+const region = IntervalBox(-5.1e2..1.4, -5.1e2..1.1)
+
+const known_roots = [
+    IntervalBox(-1.34298 ± 1e-5, -1.00134 ± 1e-5),
+    IntervalBox(-1.34030 ± 1e-5,  1.00134 ± 1e-5),
+    IntervalBox( 1.34030 ± 1e-5,  0.99866 ± 1e-5),
+    IntervalBox( 1.34298 ± 1e-5, -0.99866 ± 1e-5),
+    # following three roots are not reported accurately
```
dpsanders: Not reported accurately by the package or the original paper?

test/test_smiley.jl
```diff
@@ -0,0 +1,37 @@
...
+
+function test_all_unique(xs)
+    for x in xs
+        @test x.status == :unique
+    end
+    return nothing
+end
+
+const tol = 1e-7
+const method = Newton # NOTE: Bisection method performs badly in all examples
```
dpsanders: What do you mean by "performs badly"?

test/smiley_examples.jl
gwater: yes, i think so. as you can see the two commented roots contain values from the three following roots. i think they made an error in the first column by first skipping 0.537839 and then repeating 0.869521.

test/smiley_examples.jl
gwater: in the original paper. with the reference values i tried to stick as closely as possible to the reported data.

test/test_smiley.jl
gwater: I can't remember, either too slow or it missed roots. None of the examples worked and it seemed unhelpful to add them as @test_broken.

test/smiley_examples.jl
gwater: maybe this should say "not reported precisely" because the reported values are broadly correct, just not very precise.

lbenet:
While all tests are passing, there seems to be a conflict with current master in test/runtests.jl; can you rebase to current master?

gwater:
@lbenet ok, rebased to current master.

lbenet:
Thanks a lot; tests are passing. I am in favor of merging this.

@dpsanders Do you agree?

test/runtests.jl
```diff
@@ -6,3 +6,4 @@ using Base.Test
 include("roots.jl")
 include("newton1d.jl")
 include("quadratic.jl")
+#include("test_smiley.jl")
```
lbenet: Why not running the tests?

dpsanders: @gwater commented that one of the tests takes a very long time (40 mins). Maybe we can reduce the tolerance to get it to run faster?

lbenet: I did some tests in my laptop (using IntervalArithmetics.jl v0.13.0) and the last tests throws an error; the time it takes is quite reasonable. The error is related with the actual number of roots obtained (20 instead of 41).

I propose to have the first three tests being checked, and leave the last one commented, for the time being.

Incidentally, I also checked this against current master of IntervalArithmetics (branching from lb/iss78 in IntervalRootFinding); again the first three tests run fine and reasonably fast, but the last one throws a different error.

lbenet: To be more explicit, the test that was not passing is Example 5.5.

gwater: Hmm, I cannot reproduce the error in Example 5.5. All tests pass on my machine. Some details:
```
julia> versioninfo()
Julia Version 0.6.3
Commit d55cadc350 (2018-05-28 20:20 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)

julia> Pkg.status("IntervalArithmetic")
 - IntervalArithmetic            0.13.0
```
I've also run the tests with various tolerances: Tests pass up until a tolerance of `1e-6` above that roots start to get miscategorized or missed. At that tolerance Example 5.5 takes ca. 3 minutes on my machine which may be acceptable for the standard testing suite (esp. considering things will probably speed up more in the future).

So that's my recommendation: Raise tolerance to `1e-6` and add all examples, including Example 5.5, to the `runtests.jl` suite.

gwater: Apparently Example 5.5 fails on Travis CI, too. To be honest though, hunting this bug seems outside the scope of this pull request to me. (Clearly, the tests are uncovering an issue here but that does not mean they are broken or wrong.)

lbenet:
> Apparently Example 5.5 fails on Travis CI, too. To be honest though, hunting this bug seems outside the scope of this pull request to me. (Clearly, the tests are uncovering an issue here but that does not mean they are broken or wrong.)

I agree with you; would you mind open an issue on that? 

Regarding this PR, and precisely to go on having these more stringent tests, I propose to comment that test and merge this PR. Then, we can try to track down the problem.

Maybe @dpsanders has another idea?

dpsanders:
That sounds good to me; comment out the failing test and add an issue about it; enable the tests and merge!

lbenet:
Thanks a lot con this contribution! Merging.

Issue #78, opened 2018-06-04
Methods on IntervalBoxes should be moved to IntervalArithmetic


dpsanders:
Cross-ref: https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/158

PR #80, opened 2018-06-08
WIP: Fix to multidim Newton for new broadcasting of IntervalBox

dpsanders:
I had to jump through syntax hoops here to fix multidim Newton after the change to have `IntervalBox` contain an `SVector`.

Suggestions for changes to the syntax to make this smoother are welcome!

### DIFF
--- a/src/contractors.jl
+++ b/src/contractors.jl
@@ -1,7 +1,6 @@
 export Bisection, Newton, Krawczyk
 
 Base.isinf(X::IntervalBox) = any(isinf.(X))
-IntervalArithmetic.mid(X::IntervalBox, α) = mid.(X, α)
 
 doc"""
     Contractor{F}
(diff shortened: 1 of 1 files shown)


lbenet:
Tests are failing, because `mid` for `IntervalBox` has been removed in this PR. This is actually ok, see #78, but perhaps we need to tag a patch version in `IntervalArithmetic`, so everything is smoother, and add that version to `REQUIRE`.

dpsanders:
I think we should just have a constructor
```
IntervalBox(s::SVector)
```
for an `SVector` that contains numbers (rather than intervals);
this will simplify a bit.

dpsanders:
I think this should fix the failing tests when using IntervalArithmetic master together with  IntervalRootFinding master

lbenet:
Tests fail, though I think they are not using IntervalArithmetics/master. Locally, they get stuck.

See #83; there are some things included here that are also there.

dpsanders:
Superseded by #83.

PR #81, opened 2018-06-10
[WIP] Multidimensional Interval Newton Method to solve a system of non-linear equations

eeshan9815:
@dpsanders


eeshan9815:
This is currently an initial implementation mostly because I am having some difficulty getting `linear_hull` to work with this. Using Gaussian Elimination works fine for now.
I should be able to fix this soon, and update the PR.

PR #88, opened 2018-07-26
Add test of finding stationary points

dpsanders:


test/stationary_points.jl
```diff
@@ -0,0 +1,23 @@
...
+
+ForwardDiff.gradient(f, X::IntervalBox) = ForwardDiff.gradient(f, X.v)
+
+∇f = X -> ForwardDiff.gradient(f, X)
+
+L = 5.0
+X = IntervalBox( (-L..L+1), 2)
+
+@time rts = IntervalRootFinding.roots(∇f, X, Bisection)
+@time rts = IntervalRootFinding.roots(∇f, rts, Newton, 1e-20)
```
Kolaru: These two `@time` can be removed. I'll add them as benchmark to #105.

Also with minor adaptation Krawczyk method could be checked here too (a bit to involved change for GitHub propose change feature though).

test/stationary_points.jl
```diff
@@ -0,0 +1,23 @@
...
+
+@test length(rts) == 529
```
Kolaru:
```suggestion
@test length(rts) == 529
@test all(isunique(rts))
```

Checking that the method guarantee unicity can not hurt.

dpsanders: Good points, thanks!

Kolaru:
Superseded by #205

PR #97, opened 2018-09-29
Separated the logic of branch and bound search

Kolaru:
I finally found the courage to update to 0.7/1.0 and to implement the idea proposed by @dpsanders  in #72 to separate the logic of the search from the rest of the package. Therefore this PR superseed #72 (I do not update it though, as the scope is quite different).

Changes in this PR:
- Introduction of a generic customizable branch and bound search interface.
- Introduction of a tree type as the fundamental data structure for the search. Some advantage come from that as for example pretty printing of the tree. The `roots` function still return a simple list of intervals.
- Separation of the branch and bound logic in the `rootsearch_iterator.jl` file, so the transition to the code in another package can be done easily.
- Rework of the interface to avoid storing functions in an object (and the kind of akwardly high number of parameters that resulted).
- Use of `Root` object during the search rather than interval, which is forced by the genericity of the branch and bound algorithm (status of roots does not exactly match with the status of the elements in a search).
- Fix of #29 along the way.

PR #105, opened 2018-12-12
Add benchmarks

Kolaru:
This PR add benchmarks using `PkgBenchmark.jl`. 

These benchmarks are based on the Smiley and Chun examples used in tests and the Dietmar and Ratz example functions used in the `/perf` folder. I only added a subset of the benchmark performed in the `/perf` folder to keep the benchmark time reasonnable, while having representative benchmarks.

Of course, I will gladly add anything that seems missing.

This was tested with `PkgBenchmark.jl` `v0.2.0`.

No special infrastructure are added by this PR to automatically benchmark the package, as I don't know if it is worth for now (or will ever be).

I see this PR as usefull to the progress of #97, as it will facilitate discussion about its performance consequences.


Kolaru:
Codecov gets very confused but I don't know why. Putting back tests in their initial form does not help.

Kolaru:
Build is failing because some tests in `newton1d.jl` fail due to the latest master of `IntervalArithmetic.jl`. They are fixed in #93 though, if needed I could submit a separate hotfix though.

Kolaru:
Unless some more benchmark are requested I thing this is ready.

dpsanders:
Thanks a lot!

PR #106, opened 2019-01-14
Chebyshev_method

ys1999:
It is very much similar to Newton1d method.  But the difference lies in its convergence formula , its convergence formula contains double derivative term . Advantage of chebyshev method over newton is that chebyshev method converges more rapidly than newton method .  So I am not sure that it will be overall efficient or not??


dpsanders:
Thanks for the contribution.
For some reason the file appears as a binary and I cannot see it. Could you please try to upload it again.

Also, I suggest that you create a branch that is not `master` when you make a PR.

You will need to add some tests.

You can check performance comparisons using the `BenchmarkTools.jl` package.

ys1999:
Oops , I am sorry about this technical glitch . I will upload it again in
the way as you mentioned .


Thank You Sir

dpsanders:
Closing since not actionable as is. 
Please re-open if there is further progress.

PR #110, opened 2019-02-06
Add centered forms

dpsanders:
Add mean value forms for scalar and vector functions, and a 3rd-order Taylor form for scalar functions.


dpsanders:
The Jacobian calculation seems to be very slow for big systems.

PR #125, opened 2019-04-15
Fix doc deployment

dpsanders:


.travis.yml
```diff
@@ -25,10 +25,10 @@ notifications:
 
 after_success:
   # push coverage results to Coveralls
-  - julia -e 'cd(Pkg.dir("IntervalRootFinding")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'using Pkg; cd(Pkg.dir("IntervalRootFinding")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
```
lbenet: There is another travis configuration described in the documentation of Documenter.jl, [here](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1). It reads:
```
env:
  global:
    - DOCUMENTER_DEBUG=true

jobs:
  include:
    - stage: "Documentation"
      julia: 1.1
      os: linux
      script:
        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
        - julia --project=docs/ docs/make.jl
      after_success: skip

after_success:
  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'

sudo: false
```
I am not sure from which version of Documenter is recommended, but I suggest to use it.

PR #139, opened 2019-12-09
Add compat with IntervalArithmetic 0.16

ericphanson:
I haven't tested this, but the CI should check. It would be nice to be able to use both IntervalArithmetic 0.16 and the latest release of IntervalRootFinding.jl together.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
 ForwardDiff = "^0.10.0"
-IntervalArithmetic = "^0.15.0"
+IntervalArithmetic = "0.15, 0.16"
 Polynomials = "0.5.0"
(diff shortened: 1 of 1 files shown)


lbenet:
Thanks a lot! I was precisely dealing with problems related to this! 

To take effect, this requires a new tag; @dpsanders can you take care of this?

ericphanson:
Just to say in e.g. https://travis-ci.org/JuliaIntervals/IntervalRootFinding.jl/jobs/622651902#L219 we can see IntervalArithmetic 0.16.1 was indeed used, so the tests are indeed testing the change this PR makes (whereas if some other dependency was also holding IntervalArithmetic back to 0.15 the tests could pass but that wouldn't help us know if IntervalArithmetic 0.16.1 is OK with IntervalRootFinding).

dpsanders:
Thanks!

PR #144, opened 2020-01-13
Update Project.toml

dkarrasch:
* include newer Polynomials, StaticArrays
* remove default caret specifiers
* remove trivial (0) patch numbers


lbenet:
Thanks for your contribution and sorry for the long delay. LGTM. I'll merge it and deal with the new tag.

PR #151, opened 2020-04-21
Fixes the README example

TeroFrondelius:
You have a nonworking example. It was somewhat frustrating to discover that one needs to import some other packages to use this one.


dpsanders:
Sorry about that. Thanks for the fix!

PR #153, opened 2020-04-26
Reenable and fix example 5.5 from Smiley et al.

gwater:


gwater:
fix for #86

gwater:
unfortunately, the example takes to long to solve and some CI tests time out. So we might have to create a faster test for CI

gwater:
implements the idea described in https://github.com/JuliaArrays/StaticArrays.jl/issues/450

src/contractors.jl
```diff
@@ -20,9 +22,11 @@ end
 Multi-variable Newton operator.
 """
 function 𝒩(f::Function, jacobian::Function, X::IntervalBox, α)  # multidimensional Newton operator
-    m = IntervalBox(Interval.(mid(X, α)))
+    m = Interval.(mid(X, α))
     J = jacobian(X)
-
+    if contains_zero(det(J))
```
dpsanders, line 27: Isn't this very expensive? This should basically come out of the `\` or LU factorization.

gwater: I cant imagine the determinant is as expensive as LU - its just a fixed number of multiplicaitons and additions without any memory allocations

dpsanders: As far as I understand, the determinant does exactly an LU decomposition.

dpsanders: (ah, maybe not if this is for small StaticArrays?)

dpsanders: https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L1538

gwater: apparently StaticArrays also performs an LU decomposition for matrices larger than 4x4 https://github.com/JuliaArrays/StaticArrays.jl/blob/53b4a0ef015bd937608b57a69b6483fe13305762/src/det.jl

gwater:
Ok, given that there apparently is no way to test whether LU decomposition will work, I dont see a solution. The fundamental problem remains that IntervalArithmetic tolerates silent failures in external algorithms. I spent a few hours generalizing (and cleaning up) IntervalRootFinding to support NumberIntervals and then [the solution is easy](https://github.com/gwater/IntervalRootFinding2.jl/blob/876ba73fc77c96b90bdc8436a13f4b85a1f12ab2/src/contractors.jl#L28):
```julia
function 𝒩(f::Function, jacobian::Function, X::AbstractVector{T}, α) where T # multidimensional Newton operator
    m = T.(mid.(X, α))
    J = jacobian(X)
    try
        return convert(typeof(X), m .- (J \ f(m)))
    catch
        return X .± Inf
    end
end
```
I've tested it: The exception is triggered (disturbingly often) and eventually all roots in Smiley 5.5 are discovered.

src/contractors.jl
Kolaru: The fact that this check is even necessary is disturbing as we overload `\` for intervals with custom methods in

https://github.com/JuliaIntervals/IntervalRootFinding.jl/blob/master/src/linear_eq.jl

Either something is broken there or somehow the wrong method is picked.

dpsanders: Maybe line 165 needs to have StaticVector instead of StaticArray?

We should definitely test if this is actually getting called.

gwater: maybe it makes more sense to call gaussian_elimination explicitely (instead of the `\` operator) to make sure and obvious that no external algorithm should be called here

gwater: > The fact that this check is even necessary is disturbing as we overload `\` for intervals with custom methods in
> 
> https://github.com/JuliaIntervals/IntervalRootFinding.jl/blob/master/src/linear_eq.jl
> 
> Either something is broken there or somehow the wrong method is picked.

I can confirm that the internal method is not getting called – in fact it is broken for StaticArrays. And fixing the Array / Vector mixup is not sufficient either…

dpsanders: Are we just not actually importing \ from Base?

gwater:
ok, this version uses the internal algorithm and finally finds all the roots for Smiley 5.5

gwater:
not sure I trust the internal solver – for the matrix discussed above, it basically just returns this hard-coded value from https://github.com/JuliaIntervals/IntervalRootFinding.jl/blob/80328b460deae07f447f83c8a6dbefad21837b92/src/linear_eq.jl#L149

gwater:
> not sure I trust the internal solver – for the matrix discussed above, it basically just returns this hard-coded value from
> 
> https://github.com/JuliaIntervals/IntervalRootFinding.jl/blob/80328b460deae07f447f83c8a6dbefad21837b92/src/linear_eq.jl#L90

It's also breaking the Julia naming convention – this function is marked as mutating but it doesn't https://github.com/JuliaIntervals/IntervalRootFinding.jl/blob/80328b460deae07f447f83c8a6dbefad21837b92/src/linear_eq.jl#L102

gwater:
cleaned up the commit log a bit. this PR contains a few more changes than strictly necessary, lmk if I should separate them into other PRs

test/roots.jl
```diff
@@ -62,7 +62,7 @@ end
 
     # Bisection
     rts = roots(f, X, Bisection, 1e-3)
-    @test length(rts) == 4
```
dpsanders, line 65: I do get 4 boxes here (even though there are only 2 roots).

src/linear_eq.jl
```diff
@@ -103,10 +103,13 @@ function gauss_elimination_interval!(x::AbstractArray, A::AbstractMatrix, b::Abs
...
-    else
-        A = copy(A)
-        b = copy(b)
     end
+    _A = A
+    _b = b
+    A = similar(A)
+    b = similar(b)
+    A .= _A
+    b .= _b
```
dpsanders, line 112: Is this different / better than `_A = A; A = copy(_A)`?

gwater: If I recall correctly, calling `similar()` on a `StaticArray` gives a `MutableArray` which is required for the `setindex!()` calls below.

test/roots.jl
gwater: Yes, it gives an incorrect answer. Marking it as `@test_broken` makes transparent that `Bisection` is unreliable and probably should never be used in practise. In addition, it will notify us if `Bisection` ever starts to give the correct result here. (I admit, this is not really related to the topic of this PR–I think I just noticed this inconsistency while I was working on it.)

dpsanders: I see what you mean, but I don't agree that the result is unreliable. It is reliably giving you a set of intervals that are guaranteed to contain all of the roots in that box, and correctly telling you that it is unable to guarantee whether or not there are actually roots there.

The test is testing what the function returns, not what we would like it to ideally return ;)

gwater: Ok, that makes sense. But testing for the result 4 seems overly specific if any number above 2 would acceptable. Maybe it would be better to test that the exact solutions are contained in whatever boxes are returned? I've created a separate PR #156 for that.

dpsanders:
Thanks a lot!

PR #156, opened 2020-06-25
Make bisection test more expressive

gwater:
see our discussion in https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/153#discussion_r445127095

### DIFF
--- a/test/roots.jl
+++ b/test/roots.jl
@@ -55,6 +55,9 @@ newtonlike_methods = [Newton, Krawczyk]
 end
 
 
+in_solution_set(point, solution_intervals) =
+    any(map(interval -> point in interval, solution_intervals))
+
 @testset "2D roots" begin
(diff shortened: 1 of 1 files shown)


test/roots.jl
```diff
@@ -55,14 +55,19 @@ newtonlike_methods = [Newton, Krawczyk]
 end
 
 
+in_solution_set(point, solution_intervals) =
+    any(map(interval -> point in interval, solution_intervals))
```
dpsanders, line 59: I believe you remove the `map` and just write `any(interval -> point in interval, solution_intervals)`, which should even be slightly more efficient.

gwater: Yes, that's correct. I've adjusted the PR.

dpsanders:
Great, thanks!

PR #176, opened 2021-08-04
export interval definition for convenience

Datseris:


lucaferranti:
did you know about [ValidatedNumerics.jl](https://github.com/juliaIntervals/validatednumerics.jl) which exports all (well almost) julia intervals packages at once?

Nevertheless, I see no harm in this PR, as someone may just want to use IntervalRootFinding without e.g. ICP and TaylorModels.

Just one commment, why not using `@reexport IntervalArithmetic` instead of just re-exporting the `..`?

Datseris:
I didn't know of validatedNumerics.jl, but I do not think that would be a viable option for my case (which adds a package as a dependency to some other package), as one needs to minimize dependencies.

Sure `@reexport IntervalArithmetic` is better, I'll do that one.

Datseris:
> That is a good idea, yes. We just have to make sure that there is no problem if you also explicitly import IntervalArithmetic.jl.

Yes this is totally fine. Julia understands that the `..` that this package exports does come from IntervalArithemtic.jl and there will be no conflicts even if you import it.

Datseris:
This is good to go I think.

lucaferranti:
LGTM, thanks!

lucaferranti:
@Datseris the new release should now be available in the Julia registry.

PR #183, opened 2022-08-02
Bump version for Polynomials v3

hurak:
All tests pass, why not allowing Polynomials.jl v3 ?


vancleve:
Can we get a new version of the package that includes this commit?

Kolaru:
Of course, jut started the registration process (it will be tagged and release soon).

Issue #128, opened 2019-05-21
Do not export Contractor


dpsanders:
We should not export `Contractor` from this package; it's more important to export it from `IntervalConstraintProgramming`.

In the future this should be cleaned up (since `Contractor` in `ICP` is a specific one, whereas here it's correctly abstract).

Kolaru:
Fixed in #188

PR #188, opened 2023-03-04
Refactor contractors

Kolaru:
This PR refactor the contractors as follow
- Separate logic of contraction and refinement of roots
- Define the contractor function directly as the contractor object

This is relatively minor and should not change any results. It is however nice to have it first before I move to overhauling the interface.

### DIFF
--- a/src/contractors.jl
+++ b/src/contractors.jl
@@ -2,60 +2,6 @@ export Contractor
 export Bisection, Newton, Krawczyk
 
 """
-    𝒩(f, f′, X, α)
-
-Single-variable Newton operator.
-
-The symbol for the operator is accessed with `\\scrN<tab>`.
-"""
-function 𝒩(f, f′, X::Interval{T}, α) where {T}
-    m = Interval(mid(X, α))
-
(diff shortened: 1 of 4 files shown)


src/contractors.jl
```diff
@@ -2,125 +2,99 @@ export Contractor
...
-Multi-variable Newton operator.
+Abstract type for contractors.
 """
-function 𝒩(f::Function, jacobian::Function, X::IntervalBox, α)  # multidimensional Newton operator
-    m = Interval.(mid(X, α))
-    J = jacobian(X)
-    y = gauss_elimination_interval(J, f(m)) # J \ f(m)
-    return IntervalBox(m .- y)
-end
+abstract type Contractor{F} end
```
lucaferranti, line 9: how would you feel about renaming this to `AbstractContractor`?

Btw, while you are at it, we could close #128 and not export this?

Kolaru: I'm fine with both.

lucaferranti (approved):

Issue #180, opened 2021-11-30
Allow to unpack roots objects


Kolaru:
I think it would be nice to be able to unpack root objects as if they were tuples of `(interval, status)` e.g. have the following to work

```julia
julia> r = Root((1..2), :unique)
Root([1, 2], :unique)

julia> x, status = r
```

One interesting thing is that it would allow to iterate over the vector of `Root`s return by `roots` as follow
```julia
for (x, status) in roots(f, A)
    # do stuff
end
```

PR #190, opened 2023-05-17
Add iterate method for Root objects

edtrelo:
Solves #180

### DIFF
--- a/src/root_object.jl
+++ b/src/root_object.jl
@@ -54,6 +54,8 @@ IntervalArithmetic.isnai(r::Root) = isnai_region(root_region(r))
 function Base.:(==)(r1::Root, r2::Root)
     root_status(r1) != root_status(r2) && return false
     return isequal_region(root_region(r1), root_region(r2))
+end
+
 big(a::Root) = Root(big(a.interval), a.status)
 
 """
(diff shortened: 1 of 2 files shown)


Kolaru:
Thanks for the contribution.

If you have any idea why the x86 tests are failing, that would be great, I'm personally quite confused.

src/root_object.jl
```diff
@@ -50,3 +50,42 @@ show(io::IO, rt::Root) = print(io, "Root($(rt.interval), :$(rt.status))")
...
+    - `r`: Root object.
+    - `state`: indicates the prior element iterated.
+
+    Outputs:
+    - `r.status`: status of the root.
+    - `state`: indicates that the second iterable element of r has been returned.
+    or 
+    - nothing: the iteration has been completed.
+"""
+function Base.iterate(r::Root{T}, state::Int64) where {T}
```
lbenet, line 84: Thanks @edtrelo for your contribution.

Following on [this comment](https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/190#issuecomment-1591086103), I think the error is due to this line: it imposes `state` to be `Int64`. I think the problem is solved if `state` is required to be `Integer`

edtrelo: Thanks for the sugguestion @lbenet. I've alredy changed the type of `state`.

Kolaru:
Absolutely. I will look into it soon (tm).

Kolaru:
I've reformatted things a bit and rebased it. Thanks a lot for your work @edtrelo and sorry for the long delay

PR #194, opened 2023-09-19
update docs workflow

lucaferranti:


lucaferranti:
failure is because there are some docstrings not included in the manual

```
┌ Error: 3 docstrings not included in the manual:
│ 
│     IntervalRootFinding.guarded_derivative_midpoint :: Union{Tuple{T}, Tuple{Function, Function, Interval{T}}} where T
│     IntervalRootFinding.newton :: Union{Tuple{T}, Tuple{Function, Function, Interval{T}}, Tuple{Function, Function, Interval{T}, Int64}} where T
│     IntervalRootFinding.newton_refine :: Union{Tuple{T}, Tuple{N}, Tuple{Function, Function, Union{Interval{T}, IntervalBox{N, T}}}} where {N, T}
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in canonical @docs or @autodocs blocks.
└ @ Documenter ~/.julia/packages/Documenter/Meee1/src/utilities/utilities.jl:44
[ Info: Populate: populating indices.
```

we can relax the check or include those in the manual

Kolaru:
All those docstrings are associated to function that will be removed in #203, so I just went ahead and commented them out (not the most elegant solution, but practical).

(`newton` will only be accessible through the `roots` interface and `guarded_derivative_mid` is not necessary and was internal anyway).

Thanks a lot and sorry for the 8 months delay in this review :grimacing:

Issue #40, opened 2018-03-12
Cannot use complex functions


dpsanders:
```
julia> roots(sin, Complex(-5..5, -5..5))
ERROR: MethodError: no method matching sin(::IntervalRootFinding.Compl{IntervalArithmetic.Interval{Float64}})
Closest candidates are:
  sin(::Float64) at math.jl:419
  sin(::Float32) at math.jl:420
  sin(::Float16) at math.jl:950
```

There was an issue that meant that standard Julia complex numbers could not be used, hence the reimplementation in `complex.jl` as `Compl` that is cited here.

dpsanders:
(Though I don't see how that is being used.)

dpsanders:
After simply removing the definitions of `Compl` in `complex.jl`, this seems to work:
```
julia> roots(sin, Complex(-5..6, -5..6))
3-element Array{IntervalRootFinding.Root{Complex{IntervalArithmetic.Interval{Float64}}},1}:
 Root([3.14159, 3.1416] + [0, 0]*im, :unique)
 Root([-4.44287e-27, 3.23118e-27] + [-1.61559e-27, 1.2117e-27]*im, :unique)
 Root([-3.1416, -3.14159] + [-0, -0]*im, :unique)
```

dpsanders:
(Note that there is some weirdness with the zeros in the first and last root.)

bernardrhall:
Hi, I seem to have run into a similar problem, and I was wondering if this has been fixed?  Using the test function:

```
x=-5..5
xx = Complex(x,x)

function tst(x)
    return sin(x)
end

rts = roots(tst, xx)
```

I get the error:

> MethodError: no method matching sin(::IntervalRootFinding.Compl{Interval{Float64}})
> Closest candidates are:
>   sin(!Matched::Float16) at math.jl:1018
>   sin(!Matched::Complex{Float16}) at math.jl:1019
>   sin(!Matched::Missing) at math.jl:1070
>   ...
> 
> Stacktrace:
>  [1] tst(::IntervalRootFinding.Compl{Interval{Float64}}) at .\In[86]:6
>  [2] (::getfield(IntervalRootFinding, Symbol("#g#34")){typeof(tst)})(::IntervalBox{2,Float64}) at C:\Users\mrg\.julia\packages\IntervalRootFinding\H2vL9\src\complex.jl:34
[...]

Issue #181, opened 2022-07-17
Incorrect unique zero for 0/x + x and similar.


kemchurch:
I am computing zeroes of some pretty weird functions, and occasionally have unexpected behavior. A minimal working example (it's embedded in a lot of my functions; they depend on parameters and sometimes this kind of thing happens) is provided by the following.

```
julia> h(x) = x + zero(typeof(x))/x;

julia> r = roots(h,-1..1)
1-element Vector{Root{Interval{Float64}}}:
 Root([0, 0], :unique)

julia> h(r[1].interval)
∅
```

I think this is more of a feature of IntervalArithmetic.jl and/or ForwardDiff.jl internals than it is with IntervalRootFinding.jl, since the main reason this happens can be traced back to two things:
1. ForwardDiff.jl handles the derivative of `zero(typeof(x))/x` in such a way that `zero(typeof(x))*ForwardDiff.derivative(y->1/y,x)` is computed. This is just a hunch; I haven't looked at the internals.
3. IntervalArithmetic.jl handles certain products involving intervals with Inf and 0 in just the right way to cause trouble. Namely, `zero(typeof(x))*interval(-Inf,Inf)` will return `[0,0]`, but `interval(0)/interval(0)` returns the empty interval. This is why interval Newton converges. Despite this, our function does not have a zero, and evaluates to the empty interval there as it should.

I suspect the same thing will happen if h is replaced with `h(x) = zero(typeof(x))/x + f(x)` for any `f` that has a unique zero in [-1,1] at 0 such that `roots()` converges.

I wasn't sure where to raise this issue. One could argue that this happens because of an inconsistency with IntervalArithmetic.jl's handling of `0*interval(-Inf,Inf)` and similar. However, I noticed this problem while using IntervalRootFinding.jl, so I've raised the issue here for now.

lbenet:
Hi @kemchurch! Thanks for opening this discussion here. 

[...]

Issue #162, opened 2021-03-14
Extra roots bug for large intervals 


dpsanders:
https://discourse.julialang.org/t/question-about-interval-root-finding/57124

```jl
f(E) = tan(√(2E)) + √(E / (3π^2 - E))

roots(f, -1e100..1e100)
```

is buggy.

lucaferranti:
Also, now that I look at the function better `E=0` should also be a root, but that goes undetected with Newton :thinking:. Krawczyk and Bisection can find it at the cost of a couple of extra roots. 

```julia-repl
julia> roots(f, 0..30, Bisection)
5-element Vector{Root{Interval{Float64}}}:
 Root(Interval(1.2337005023828356, 1.233700557778276), :unknown)
 Root(Interval(15.064561967948055, 15.064562025101973), :unknown)
 Root(Interval(3.844642092262015, 3.8446421485298243), :unknown)
 Root(Interval(11.103304945220835, 11.103305008980778), :unknown)
 Root(Interval(0.0, 8.972354260755779e-8), :unknown)

julia> roots(f, 0..30, Krawczyk)
5-element Vector{Root{Interval{Float64}}}:
 Root(Interval(15.064562002776027, 15.064562002781207), :unique)
 Root(Interval(11.103304945220835, 11.103305008980778), :unknown)
 Root(Interval(1.2337005023828356, 1.233700557778276), :unknown)
 Root(Interval(3.8446421166049274, 3.844642116613573), :unique)
 Root(Interval(0.0, 8.972354260755779e-8), :unknown)
```

Also, those extra intervals contain vertical asymptotes of the function (pi^2/8 and 9/8*pi^2), which I guess explains why the algorithm cannot through it away.

dpsanders:
The root at 0 cannot be proved because there is not an open interval around it, since `sqrt` is undefined for negative numbers.

dpsanders:
But the root at 0 should still be reported as `:unknown`. That seems like a serious bug! I am wondering if that is what is giving the empty interval.

lucaferranti:
yeah I also wondered whether there was a relation between the returned empty set and the root at zero. It seems that starting with "small" intervals in the form `[0, a]` the root `0` is correctly found as `:unknown` , for bigger intervals the root at zero remains undetected. Furthermore, if 0 is in the interior of the starting interval, then sometimes an empty set is returned.

```
[...]

Issue #157, opened 2020-06-26
Allow roots(f, Xs) for a vector Xs


dpsanders:
This currently fails

dpsanders:
Duplicate of #117

PR #203, opened 2024-05-29
Compatibility with IntervalArithmetic v0.22

Kolaru:
This PR aims at providing compatibility with IntervalArithmetic v0.22, which is breaking.

The user facing changes are
- Decorated intervals must be used (which is the default `Interval` in the latest releases of IntervalArithmetic)
- The signature of the `roots` function changed to `roots(f::Function, X::Union{Interval, AbstractVector} ; kwargs...)`, with the following consequences
    - Manual derivatives and contractors must now always be passed as keyword arguments. This greatly simplify the internal logic of the `roots` function.
    - No more `IntervalBox`. Multidimensional problem are specified by returning a vector of intervals, and giving a vector of intervals as initial search region.
    - Normal vectors of intervals are supported. They are significantly (3x) time slower for a simple 2D problem than SVector, but more convenient.
- Features of the packages outside of the `roots` function are unsupported for the time being (e.g. `Slopes` and quadratic equations). If you were using them, please open an issue.
- `roots` works with `@exact`.

This PR does not yet contain an update of the documentation, but is otherwise ready for review (if the tests pass on CI ^^').


OlivierHnt:
You are mentioning a slowdown, this is only due to using `Vector` instead of `SVector`?

Also, does this PR address some of the following issues #41, #52, #77, #98, #107, #117, #124, #162, #181?

What is the status of #40 about complex intervals?

Kolaru:
It fixes or superseeds #41 #52 #77 #107 #157 
and also fixes without yet include tests #98 #162 #181

It half-fixes #40 as complex intervals are supported, but require explicit derivative.

It does nothing about #124 and propably make #117 a bit worse (`roots.(f, Xs)` should be used in that case).

OlivierHnt:
Out of curiosity, why is the keyword argument of `roots` called `contractor`? Why not `alg` for "algorithm"?

Also, a lot of examples in the docs you added have the NG flag, do you think it is worth using `@exact` to prevent them? But maybe that is more confusing.

Kolaru:
> Out of curiosity, why is the keyword argument of `roots` called `contractor`? Why not `alg` for "algorithm"

`contractor` is the most technically correct term, also according to the literature.

The algorithm is branch-and-prune, the contractor is only a small part of it. That being said, I am fine with any of `contractor`, `algorithm`, or `method`.

The doc should anyway be extended to discuss what's the difference between Newton and Krawczyk.

> Also, a lot of examples in the docs you added have the NG flag, do you think it is worth using `@exact` to prevent them? But maybe that is more confusing.

I plan to discuss that separately. The `_com_NG` flag is purely related to IA.jl, and I think it is fair to assume users will either accept it, or know it from using IA in the first place.

I have been considering not showing the decoration or flag of the interval in the roots. Instead, we could add a flag to the root object, when the assumption of the method are not fullfileld (i.e. the root interval has the `NG` flag, a decoration below `dac`, or even when the derivative doesn't have at least the `dac` decoration).

OlivierHnt:
> To be fair, I have been considering not showing the decoration or flag of the interval in the roots. Instead, we could add a flag to the root object, when the assumption of the method are not fullfileld (i.e. the root interval has the NG flag, a decoration below dac, or even when the derivative doesn't have at least the dac decoration).

Do you think you can loose information by having a single global decoration and NG flag? For instance, in one of the example some roots have the "NG" flag and others don't.

Kolaru:
In my idea, the information would be kept on the individual intervals, just not displayed. I assume that at a first glance, what's important is if the root is valid as a whole.

I'll keep that for another PR though, this one has already enough changes.

Kolaru:
With the update to the doc, this PR is not fully ready for review.

OlivierHnt:
Ok! I can help with the review once you decide your PR is ready.

schillic:
> With the update to the doc, this PR is not fully ready for review.

@Kolaru was the "not" a "now" with a typo? :)

OlivierHnt:
Yes I believe that was the case 🙂

src/region.jl
```diff
@@ -0,0 +1,50 @@
+in_region(x, Y::Interval) = in_interval(x, Y)
+in_region(x::AbstractVector, Y::AbstractVector{<:Interval}) = all(in_interval.(x, Y))
+
+function intersect_region(X::Interval, Y::Interval)
+    intersection = intersect_interval(bareinterval(X), bareinterval(Y))
+    dec = min(decoration(X), decoration(Y))
+    guarantee = isguaranteed(X) && isguaranteed(Y)
+    return IntervalArithmetic._unsafe_interval(intersection, dec, guarantee)
+end
```
OlivierHnt, line 9: I think this may be simplified to

```Julia
function intersect_region(X::Interval, Y::Interval)
    dec = min(decoration(X), decoration(Y))
    return IntervalArithmetic.setdecoration(intersect_interval(X, Y), dec)
end
```

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+
+intersect_region(X::AbstractVector, Y::AbstractVector) = intersect_region.(X, Y)
+
+isempty_region(X::Interval) = isempty_interval(X)
+isempty_region(X::AbstractVector) = any(isempty_region.(X))
```
OlivierHnt, line 14: To avoid allocations `any(isempty_region, X)`?

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+
+isequal_region(X::Interval, Y::Interval) = isequal_interval(X, Y)
+isequal_region(X::AbstractVector, Y::AbstractVector) = all(isequal_region.(X, Y))
```
OlivierHnt, line 17: Similarly, `all(t -> isequal_region(t...), zip(X, Y))`?

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+
+isbounded_region(X::Interval) = isbounded(X)
+isbounded_region(X::AbstractVector) = all(isbounded.(X))
```
OlivierHnt, line 20: Similarly, `all(isbounded, X)`?

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+
+isnai_region(X::Interval) = isnai(X)
+isnai_region(X::AbstractVector) = any(isnai.(X))
```
OlivierHnt, line 23: `any(isnai, X)`?

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+
+diam_region(X::Interval) = diam(X)
+diam_region(X::AbstractVector) = maximum(diam.(X))
```
OlivierHnt, line 26: `maximum(diam, X)`?

src/region.jl
```diff
@@ -0,0 +1,50 @@
...
+function bisect_region(X::SVector{N, T}, α) where {N, T}
+    i = argmax(diam.(X))
+    x1, x2 = bisect_region(X[i], α)
+    X1 = SVector{N, T}(k == i ? x1 : X[k] for k in 1:N)
+    X2 = SVector{N, T}(k == i ? x2 : X[k] for k in 1:N)
+    return X1, X2
+end
+
+istrivial(X::Interval) = decoration(X) <= trv
+istrivial(X::AbstractVector) = any(istrivial.(X))
```
OlivierHnt, line 50: `any(istrivial, X)`?

src/roots.jl
```diff
@@ -1,24 +1,84 @@
...
-struct RootProblem{T}
+struct RootProblem{C, F, G, R, S, T}
+    contractor::Type{C}
+    f::F
+    derivative::G
+    region::R
+    search_order::Type{S}
     abstol::T
+    reltol::T  # TODO
+    max_iteration::Int  # TODO
```
OlivierHnt, line 14: What is the meaning of "TODO"?

OlivierHnt:
I went through the changes. Since I am not familiar with this library, I cannot comment on the some inner changes, but in terms of IA v0.22, it is looking good!

I think the breaking changes are consistent with IA changes, so it may be good to merge this PR soon-ish.

src/region.jl
Kolaru: It can, indeed.

src/roots.jl
Kolaru: They options are straight up not used currently 😬.

I've moved this to the docstrings.

Kolaru:
Thanks a lot ! I have added two missing docstrings.

I will go through the doc to make sure everything is described, and then I think it would be indeed ready to be merged.

Kolaru:
Surprising myself, I had already updated the doc and added an entry in NEWS.md, so I am merging.

PR #214, opened 2025-03-18
Update workflows

abhro:
- Use dependabot to track action dependencies
- Update action versions
- Use external actions like julia-actions/cache and julia-actions/julia-docdeploy to simplify workflow files


Kolaru:
Sorry for the late reaction, and thanks a lot for the update!

PR #233, opened 2026-01-28
Implement continuing processing on error

### LINKED ISSUE #232: Unexpected library error message when finding roots of simple function
I was testing interval root finding to see if it would work for my application and created this simple problem:
```julia
using IntervalArithmetic, IntervalRootFinding
using LinearAlgebra: norm
using StaticArrays: SVector
using IntervalArithmetic.Symbols
using Plots

function one_root()
    p(t, p0::AbstractVector, d::AbstractVector) = p0 .+ d .* t

    f(t) = norm(p(t, SVector(-2.0, 0.0, 0.0), SVector(1.0, 0.0, 0.0))) - 1.0

    plot(f, 0.0:0.01:4.0)
    roots(f, 0.0 .. 10)
end
```
When I run `one_root` I get this error:

`ERROR: ArgumentError: `<` is purposely not supported for overlapping intervals. See instead strictprecedes` 

f(t) is not that weird a function. It has a zero at 1 and at 3:

Image

Am I misusing the `IntervalRootFinding` library? Or is this a bug?

Kolaru:
Fix #232 

By default, with this PR, errors are ignored, and the corresponding regions are bisected further, in the hope that by reducing them, the error will go away.

This can be disable by setting the new option `bisect_on_error` to false.

Whether an error occurred, and why does a root stop being processed are both stored in the Root object, which allows to give more information at the end of the root search.

With this PR, the following works:

```julia
using IntervalArithmetic, IntervalRootFinding
using LinearAlgebra
using StaticArrays
using IntervalArithmetic.Symbols

function one_root()
    p(t, p0::AbstractVector, d::AbstractVector) = p0 .+ d .* t

    f(t) = norm(p(t, SVector(-2.0, 0.0, 0.0), SVector(1.0, 0.0, 0.0))) - 1.0

    roots(f, 0.0 .. 10)
end
```

Resulting in
```julia
julia> one_root()
3-element Vector{Root{Interval{Float64}}}:
 Root([1.0, 1.0]_com_NG, :unique)
 Root([2.0, 2.0]_com, :unknown)
    Not converged: region size smaller than the tolerance
    Warning: an error was encountered during computation
 Root([3.0, 3.0]_com_NG, :unique)
```

Doc and tests are missing and will come soon.

@brianguenter Does it look like a fine solution to your issue?

### DIFF
--- a/src/roots.jl
+++ b/src/roots.jl
@@ -12,6 +12,25 @@ struct RootProblem{C, F, G, R, S, T}
     max_iteration::Int
     where_bisect::T
     bisect_on_error::Bool
+
+    function RootProblem(
+            contractor::Type{C},
+            f::F,
+            derivative::G,
+            initial_root::R,
+            search_order::Type{S},
+            abstol::T,
+            reltol::T,
+            max_iteration::Int,
+            where_bisect::T,
+            bisect_on_error::Bool) where {C, F, G, R, S, T}
+
+        X = root_region(initial_root)
+        size(X) != size(f(X)) && throw(DimensionMismatch("input and output dimensions of f must be the same"))
+        return new{C, F, G, R, S, T}(
+            contractor, f, derivative, initial_root, search_order, abstol, reltol,
+            max_iteration, where_bisect, bisect_on_error)
+    end
 end
 
 """
--- a/src/roots.jl
+++ b/src/roots.jl
@@ -13,24 +13,24 @@ struct RootProblem{C, F, G, R, S, T}
     where_bisect::T
     bisect_on_error::Bool
 
-    function RootProblem(
-            contractor::Type{C},
-            f::F,
-            derivative::G,
-            initial_root::R,
-            search_order::Type{S},
-            abstol::T,
-            reltol::T,
-            max_iteration::Int,
-            where_bisect::T,
-            bisect_on_error::Bool) where {C, F, G, R, S, T}
+    # function RootProblem(
+    #         contractor::Type{C},
+    #         f::F,
+    #         derivative::G,
+    #         initial_root::R,
+    #         search_order::Type{S},
+    #         abstol::T,
+    #         reltol::T,
+    #         max_iteration::Int,
+    #         where_bisect::T,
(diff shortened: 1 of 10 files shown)


brianguenter:
this looks good to me.

Kolaru:
@OlivierHnt The more I think about it, the more I think only the errors coming from IntervalArithmetic.jl really have a chance to be relevant here. So I would put this on hold, until one of us implement the new error type there. What do you think?

On a side note, there are more changes that necessary, because I also spent some time fixing the doctests while I was at it, even so it is not the core of this PR.

OlivierHnt:
Sounds good, it's better to properly resolve the issue instead of making a temporary "hotfix".
I can open the PR for the error type in IntervalArithmetic.

src/roots.jl
```diff
@@ -44,11 +45,14 @@ Parameters
...
     Default: `0.0`.
 - `max_iteration`: The maximum number of iteration, which also corresponds to
-    the maximum number of bisections allowed. Default: `typemax(Int)`.
+    the maximum number of bisections allowed. Default: `100_000`.
 - `where_bisect`: Value used to bisect the region. It is used to avoid
     bisecting exactly on zero when starting with symmetrical regions,
     often leading to having a solution directly on the boundary of a region,
     which prevent the contractor to prove it's unicity. Default: `127/256`.
+-`bisect_on_error`: Whether a region that errors when the function is applied to
+    to it should be bisected. If false, when an error happen, the root search is
```
OlivierHnt: Looks like there is a typo "to" is repeated twice

OlivierHnt:
I am wondering if the default should be `true` for `bisect_on_error`...

Kolaru:
@OlivierHnt I've added support for `InconclusiveBooleanOperation`.

Also now, instead of having a flag to ignore errors or not, `roots` take a list of error types and ignores only those specified (default to only `InconclusiveBooleanOperation`).

What do you think about this design?

OlivierHnt:
Sounds reasonable that the error is not being thrown but it is still recorded in the output roots.

Issue #142, opened 2019-12-24
Difficult benchmark problem


dpsanders:
https://link.springer.com/article/10.1023/A:1008236911603

dpsanders:
As stated in Ratschek 1993, the last equation should be `x1 * x3 - x2 * x4 == 0`, not `x1 - x3 - x2*x4 == 0`.

PR #239, opened 2026-04-19
Infer root type from the provided function

Kolaru:
This PR changes how the internal root type is determine:

Now it uses the return type of the function, regardless of the type of the initial region given. This should avoid unnecessary conversions.


Kolaru:
Benchmarks are slower due to the inference only, and fail because I had to change them for the future.

A longer benchmark (few seconds to a minute) would be nice to be less sensitive to that kind of short setup effects (see #142 ).

Issue #131, opened 2019-07-09
Reciprocal of reciprocal misses root


jwscook:
(Unnecessarily) taking the reciprocal of a reciprocal causes the package to miss a root:

```
julia> using IntervalArithmetic, IntervalRootFinding

julia> f(x) = (1-x)*(1+x)
f (generic function with 1 method)

julia> roots(x->f(x), interval(-2, 2))
2-element Array{Root{Interval{Float64}},1}:
 Root([0.999999, 1.00001], :unique)  
 Root([-1.00001, -0.999999], :unique)

julia> roots(x->1/(1/f(x)), interval(-2, 2))
1-element Array{Root{Interval{Float64}},1}:
 Root([-1.00001, -0.999999], :unknown)
```

Is this behaviour expected?

```
  [d1acc4aa] IntervalArithmetic v0.15.0
  [d2bf35a9] IntervalRootFinding v0.4.0
  [8197267c] IntervalSets v0.3.1
  [d8418881] Intervals v0.3.3
```

```
Version 1.0.4-pre.0 (2018-12-19)
release-1.0/5b7e8d9d4e (fork: 362 commits, 334 days)
```

dpsanders:
`1 / f(x)` has poles (infinities) at plus and minus 1, so `1 / (1 / f(x))` is not really even defined there?

dpsanders:
But maybe it's true that it should catch that "something funny happens" at each pole.

dpsanders:
Actually on Julia 1.1 I get
```
julia> roots(x -> 1 / ( 1 / f(x)), -10..10)
3-element Array{Root{Interval{Float64}},1}:
 Root([0.999999, 1.00001], :unknown)
 Root([-1, -0.999999], :unknown)
 Root([-1.00001, -1], :unknown)
```

dpsanders:
(With master (?) branches of the interval packages, if that makes a difference.)

Kolaru:
The expected result is indeed to have the roots status to be `:unkown`.

With the stable versions of the packages (`IntervalArithmetic 0.15.2` and `IntervalRootFinding 0.4.0`) and Julia 1.1, I get correct result too.

Anyway, it would not hurt to have this in the test suite.

jwscook:
Ok, great!

dpsanders:
Can you confirm that the behaviour is correct with Julia 1.1? Is there a reason you reopened the issue?

jwscook:
[...]

Issue #99, opened 2018-10-04
Krawczyk fails for functions with interval parameters


Kolaru:
```jl
julia> a = 2..2.1
[2, 2.10001]

julia> roots(x -> x^2 - a, -4..4, Krawczyk)
0-element Array{Root{Interval{Float64}},1}
```

Newton seems to be fine

```jl
julia> roots(x -> x^2 - a, -4..4, Newton)
2-element Array{Root{Interval{Float64}},1}:
 Root([1.4141, 1.44946], :unique)
 Root([-1.44946, -1.4141], :unique)
```

dpsanders:
Thanks for the report.

Krawczyk hasn't received much attention. Presumably it's *supposed* to work for cases like this?

Bisection also does badly, since it ends up bisecting the interval that is the solution :/

Kolaru:
The problem is not the interval parameters, but the fact that our favorite example currently fails with Krawczyk, because the derivative is 0 at the middle point:

```jl
julia> roots(x -> x^2 - 2, -4..4, Krawczyk)
0-element Array{Root{Interval{Float64}},1}
```

The reason is given in the new issue #102.

If the interval is not symmetric, it does just fine:

```jl
julia> roots(x -> x^2 - (2..2.1), -4..4.0001, Krawczyk)
2-element Array{Root{Interval{Float64}},1}:
 Root([1.4141, 1.44947], :unique)
 Root([-1.44947, -1.4141], :unique)
```

Theoretically it is supposed to work fine in this case because for an interval `A`, the interval extension of `x^2 - A` is also a (bad) interval extension for `x^2 - a` for any `a` in `A`. Then, all methods may bisect the solution if they are unable to conclude that it exists for all `a` in `A`. In that case it is the parameter interval `A` that should be bisected.

On the up side, I need to do this, so I will be implementing an algorithm doing exactly this tonight :D

dpsanders:
This seems to be fixed now:
```
julia> a = 2..2.1
[2, 2.10001]

julia> roots(x -> x^2 - a, -4..4, Krawczyk)
2-element Array{Root{Interval{Float64}},1}:
 Root([1.41409, 1.44947], :unique)
 Root([-1.44946, -1.4141], :unique)
```

dpsanders:
Needs a test.

Kolaru:
I've added it to the (ever growing) list of missing tests in #115.

PR #243, opened 2026-04-29
Complete tests

### LINKED ISSUE #115: Complete tests suites
While testing #114 I noticed that a problem was caught in the wrong test suite, namely `NaN return value` rather than `Out of domain`, because the latter does not test multidimensional function. This made me realize that several test suites are incomplete.

So I open this issue to list missing tests. I'll add them in a future PR.

Currently the todo list is:

- [ ] Add multi dimensional tests to the `Out of domain` suite.
- [ ] Add multi dimensional tests to the `Infinite domain` suite.
- [ ] Add one dimensional tests to the `NaN return value` suite.
- [ ] Use test in `test/bisection.jl`.
- [ ] Adapt tests in `test/findroots.jl` to the `roots` suite.
- [ ] Add tests in `test/newton1d.jl` to the tests for the 1D `roots` suite.
- [ ] Test that the examples in the `example` folder run without error.
- [ ] Add some tests from the sources given in `docs/test_suites.md`.
- [ ] Test function with interval parameter (cf. #99).
- [ ] Add the convoluted double reciprocal case from #131.

Please let me know here if other test suites need to be completed.

Kolaru:
Fix #115

Also suppresses warning from tests that printed them.

This is not finished yet, I open it however to allow discussion on one weird behavior that I would while completing the tests:

`bisect` use the decoration of the parent interval. This is very reasonnable in general, however it also means that interval bisected from `interval(-Inf, Inf)` will all have the decroation `dac`. I belive that this is not what we want here, so I changed it.

Reporting the list from #115

- [x] Add multi dimensional tests to the `Out of domain` suite.
- [x] Add multi dimensional tests to the `Infinite domain` suite.
- [x] Add one dimensional tests to the `NaN return value` suite.
- [x] Use test in `test/bisection.jl`.
- [x] Adapt tests in `test/findroots.jl` to the `roots` suite.
- [x] Add tests in `test/newton1d.jl` to the tests for the 1D `roots` suite.
- [x] Test that the examples in the `example` folder run without error.
- [ ] Add some tests from the sources given in `docs/test_suites.md`.
- [x] Test function with interval parameter (cf. #99).
- [x] Add the convoluted double reciprocal case from #131.


OlivierHnt:
2 cents thought on the *intempestive* `dac` decoration:

The `dac` decoration tells us that the intervals containing the roots originate from an unbounded domain. Isn’t that actually a feature?
In other words, should we really be striving for `com` here? A bounded interval carrying the `dac` decoration unambiguously reflects the "provenance of the computation", no?

The main concern (or worry, so to speak) is that decorations always feel inherently problem-dependent to me. What counts as a 'safe' decoration really depends on the user's context, doesn't it? (in particular, this is why all set operations result in the `trv` decoration as per the IEEE standard).

If so, following that logic, users should perhaps upgrade the decoration to `com` themselves afterwards:

```
interval_with_root = interval(1, 2, dac)
IntervalArithmetic.setdecoration(interval_with_root, com) # [1.0, 2.0]_com
```

EDIT: Or there could be a special keyword `root(...; neglect_dec = (:dac, :def), ...)` (in the same vein of what you did for set operations in IntervalArithmetic)?

Kolaru:
After sleeping on it, I think that I agree, we should just keep the decoration while bisecting, because the interval *may* come from a previous computation, that got the `dac` decoration for other reasons.

I'm considering hiding the decoration while displaying the root however, but that should be done in a different PR.

Also, I will open a PR in IA.jl to have `Decoration` be public, and both `decoration` and `Decoration` have clear docstrings stating that decoration `dac` and `com` are both in principle totally fine, and that even `def` is okay in most cases.

PR #245, opened 2026-05-11
Prepare v0.7 release

Kolaru:
This PR introduces the v0.7 release.

It is expected to be the last release before version v1.0. Therefore all downstream packages are invited to test it and share opinions before v0.7 is released.

I will make an announcement in the Slack channel at least once this is merged.

@dpsanders @lbenet @OlivierHnt


OlivierHnt:
Fire away 🚀

Kolaru:
Changing the default is making the doctests fail (as always), I'll fix that first :)

Kolaru:
The doctests should now be fixed, but I notice #246 in the process.

PR #251, opened 2026-06-23
Use ` .. ` (with spaces)

lbenet:
It seems that Julia nightly gets sometimes confussed with the use of `..` to represent intervals; see e.g. [this error](https://github.com/JuliaIntervals/TaylorModels.jl/actions/runs/28050996276/job/83041262186#step:6:426); this prevents a succesful precompilation. This PR aims to solve that, surounding it by spaces, ` .. `. I though i have done it before...


lbenet:
Merging, as agreed in the JuliaInterval call

dpsanders:
Julia bug in parsing of `..` reported [here](https://github.com/JuliaLang/julia/issues/62200).

dpsanders:
@lbenet This should have been fixed in https://github.com/JuliaLang/julia/pull/62206/

